PR #58, opened 2022-06-13
change cron time to every 12 hours

### LINKED ISSUE #57: Changing cron time to 7:00 PM
For checking the progress of the students for the entire day we can update the data at 7:00 PM just before EOD rather than in the morning.

kunatastic:
fixes #57

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,7 @@ name: Main
 
 on:
   schedule:
-    - cron: "30 13 * * *" # 7:00PM IST
+    - cron: "0 */12 * * *" # every 12 hours
 jobs:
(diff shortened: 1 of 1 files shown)


bodhish:
Can we run it twice a day?

kunatastic:
Yeah!! we can do that we will be even better

bodhish (approved):

bodhish:
I thought the data loading service handled it. We should fix it.

rithviknishad:
@bodhi nope, the scrapper seems to simply be using start and end date with no time specified. I'll try to fix it ASAP, but need to discuss abt how to remove duplicates in already generated data in today's EOD.

PR #80, opened 2022-09-03
Improve info card hover effect

rabilrbl:
**Changes**
- New hover effect
- Add link around profile image

**Screenshots**
| Before | After |
| - | - |
|image|image|


gigincg:
@rabilrbl The same component seems to be used over here

image

Let's not have the hover effect here

PR #89, opened 2022-09-27
Added badge glow and sparkle.

skks1212:
Redesign to the badges view in learning dashboard.

https://user-images.githubusercontent.com/23238460/192822929-761651e6-ede6-4993-8932-5bdd0e581136.mp4


components/Sparkles.jsx
```diff
@@ -0,0 +1,7 @@
+export default function Sparkle(props) {
+    return (
+        <div className={"sparkle"} style={props.style}>
```
khavinshankar, line 3:
```suggestion
        <div className="sparkle" style={props.style}>
```

components/contributors/BadgeIcons.jsx
```diff
@@ -19,15 +20,32 @@ function useOnClickOutside(ref, handler) {
...
+      break;
+
+    case 1:
+      glow = "glow-silver";
+      break;
+
+    default:
+      break;
+  }
+
```
khavinshankar, line 42: this would be better as a function, some thing like
```suggestion
 const glow = () => {
    const currentLevel = skill.levels.indexOf(skill.currentLevel) + 1;

    switch (skill.levels.length - currentLevel) {
      case 0:
        return "glow-gold";
  
      case 1:
        return "glow-silver";
  
      default:
        return "";
    }
 }

```

khavinshankar (approved):

PR #98, opened 2022-10-22
sort by points if category points same

### LINKED ISSUE #96: discrepancy in top contributors data
image

Ritesh-Aggarwal:
closes #96

### DIFF
--- a/pages/leaderboard.js
+++ b/pages/leaderboard.js
@@ -51,8 +51,10 @@ export default function Home(props) {
       );
     }
 
-    filteredContributors = filteredContributors.sort(
-      (a, b) => a.weekSummary[sortBy] - b.weekSummary[sortBy]
+    filteredContributors = filteredContributors.sort((a, b) =>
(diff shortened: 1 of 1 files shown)


rithviknishad (changes requested):
@Ritesh-Aggarwal can we do this sorting inside getStaticProps()?

The order wouldn't change for a given day right, so let the server side do the computation and keep the client-side light.

Ritesh-Aggarwal:
> @Ritesh-Aggarwal can we do this sorting inside getStaticProps()?
> 
> The order wouldn't change for a given day right, so let the server side do the computation and keep the client-side light.

I don't think that would improve performance as we have sorting/filtering parameters like show/hide core members, and many more in client side, so we need to have this logic in client side itself.
If we pre-calculate all the sorted arrays for every condition it may increase the size of page and nextjs recommends only passing required data in `getStaticProps()`
https://nextjs.org/docs/messages/large-page-data

rithviknishad (approved):

PR #131, opened 2023-09-22
Fix theme button in header

### LINKED ISSUE #130: Darkmode Switch Badge default should be dark
image

Currently the app's default theme is dark mode, but the switch/selector's default state is light mode. This means that when you've never switched the Selector, the app would be in Dark mode and the selector would claim to be on light mode

Also use a button instead of a switch for Dark and Light mode similar to https://windmill-dashboard.vercel.app/

Pranshu1902:
Fixes #130 
Removed the switch and used button instead. Made the default theme as dark
Dark:
image

Light:
image

### DIFF
--- a/components/ThemeSwitch.tsx
+++ b/components/ThemeSwitch.tsx
@@ -1,6 +1,5 @@
 "use client";
 
-import { Switch } from "@headlessui/react";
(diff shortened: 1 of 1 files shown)


rithviknishad (approved):
Nice work!

PR #167, opened 2023-12-13
added turnaround time for merged prs

khavinshankar:
image

image

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -68,6 +68,7 @@ export interface Activity {
   link: string;
   text: string;
   collaborated_with?: string[];
+  turnaround_time?: number;
 }
 
 export interface OpenPr {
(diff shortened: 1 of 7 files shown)


khavinshankar:
@rithviknishad 

Currently, there is no way to identify the issue number if the issue is linked (connected) using UI, so lets handle that part while moving to Graphql.

This is the event created for connection using UI
image

If we want to do this now using rest API, we need to iterate through all the open issues and get the cross-referenced event which seems unnecessarily expensive.

PS: This is an issue only when the issue is connected with PR in PR's UI and not vice versa.

rithviknishad (approved):

lib/types.ts
```diff
@@ -68,6 +68,7 @@ export interface Activity {
   link: string;
   text: string;
   collaborated_with?: string[];
+  turnaround_time?: number;
```
rithviknishad, line 71: There's a typo btw 😅 

`turnaround_time` here, but generated data has `turn_around_time`

rithviknishad: image

rithviknishad (approved):

PR #198, opened 2024-01-02
fix review count

### LINKED ISSUE #105: Bug in the reviews count
## Bug

From November 1st to November 11th, I've reviewed only 11 PR's so far.
However, in the leaderboard stats, it says I've made 38 reviews.

This might be because in a single PR multiple follow-up reviews or multiple different line suggestions were made.

image

## Possible Solution
Let the scrapper remain the same as the individual review is required for the timeline. However, when counting we could squash multiple reviews of the same PR into one.

cc: @gigincg @bodhish

konavivekramakrishna:
Fixes #105 

Before:
image

After:
image


konavivekramakrishna:
@rithviknishad 

in timeline if there are multiple review for same pr it should remain same, just we have to fix the count right?image

konavivekramakrishna:
@rithviknishad 

could you please review it

lib/api.ts
```diff
@@ -191,10 +197,27 @@ export function getCalendarData(activity: Activity[]) {
...
       acc[date].count += 1;
+
+      const uniquePrReviews = new Set();
+
       if (acc[date][activity.type]) {
-        acc[date][activity.type] += 1;
+        if (
+          activity.type != "pr_reviewed" ||
+          (activity.type == "pr_reviewed" &&
+            uniquePrReviews.has(activity.title))
```
konavivekramakrishna, line 207: this has to be 
` !uniquePrReviews.has(activity.title))`
the condition should not be true
but keeping not true leads incorrect not sure why
but removing the not true gives correct pr_reviews values, even i checked manually

lib/api.ts
```diff
@@ -100,9 +100,14 @@ export function getContributorBySlug(file: string, detail = false) {
     ...activityData,
     activity: [...activityData.activity, ...getSlackMessages(data.slack)],
   };
+  let AllUnqPrReviews = new Set();
```
rithviknishad, line 103: Keeping this set as a global state could lead to inconsistent results. Prefer keeping it localized specific to a contributor.

lib/api.ts
```diff
@@ -100,9 +100,14 @@ export function getContributorBySlug(file: string, detail = false) {
...
 
-  const weightedActivity = activityData.activity.reduce(
+  let weightedActivity = activityData.activity.reduce(
     (acc, activity) => {
+      if (activity.type == "pr_reviewed") {
+        AllUnqPrReviews.add(activity.title);
```
rithviknishad, line 108: is there no `id` instead of title?

lib/api.ts
```diff
@@ -191,10 +197,27 @@ export function getCalendarData(activity: Activity[]) {
         };
       }
       acc[date].count += 1;
+
+      const uniquePrReviews = new Set();
```
rithviknishad, line 201: Could you explain why there are Two sets? All Unique and Unique? Could you explain the algorithm for how you are solving the problem?

konavivekramakrishna: The AllUnq set is used to count the total unique pull requests (PRs) reviewed.

and the 2nd set is For PRs reviewed in the last 7 days, we calculate and count the reviews for each date (day) and sum them up over this period.

In this process, if a PR has multiple reviews on the same day or on subsequent days, it is only counted once for its first review. Any subsequent reviews of the same PR are not counted again.

lib/api.ts
konavivekramakrishna: image
no

lib/api.ts
konavivekramakrishna: I don't think it causes inconsistent results, because for each contributor we are calculating separately.

and I didn't get how to keep it localized specific to a contributor.

lib/api.ts
konavivekramakrishna: image

yes

rithviknishad:
image

lib/api.ts
rithviknishad: @konavivekramakrishna multiple distinct reviews on the same PR must be counted. But we shouldn't count the individual review comment on the PR. We should only count one review event.

rithviknishad:
@konavivekramakrishna conversation https://github.com/coronasafe/leaderboard/pull/198#discussion_r1472745737 has not been resolved

lib/api.ts
konavivekramakrishna: @rithviknishad 
For example, in the following pr: https://github.com/coronasafe/care_fe/pull/6479

what should be the count of pr_reviewed

The count of 'pr_reviewed' for this pull request in the current staging leaderboard is 13

konavivekramakrishna: > @konavivekramakrishna multiple distinct reviews on the same PR must be counted. But we shouldn't count the individual review comment on the PR. We should only count one review event.

@rithviknishad are you referring to multiple individuals reviewing the same pull request, or do you mean counting multiple distinct reviews by a single person?"

rithviknishad: No. For example, the recent review I've made on your PR in care (https://github.com/coronasafe/care_fe/pull/7152#pullrequestreview-1858666162), I've only requested changes ONCE (and you might have also got a mail notification once for the same). However, leaderboard shows 2 review events even though I've made only 1 review event.

image


image

rithviknishad: The review event seems to be having an ID. Could you update the types? Or if we are not scrapping the ID, can we adjust the scraper to report only one review event for an ID?

rithviknishad:
@konavivekramakrishna any updates on this?

rithviknishad:
Closing this PR due to no activity

PR #207, opened 2024-01-24
Add CSS tooltip styling. 

### LINKED ISSUE #175: Migrate `Tooltip` jsx component to use CSS based tooltip to reduce client side javascript
Currently `<Tooltip />` component is a component under `"use client"` and works using client-side javascript.

### Tasks

- [ ] Tooltip can be implemented without using javascript, use CSS instead.
- [ ] Remove `"use client"` directive since it no longer requires client interactivity (JS)

AshrafMd-1:
Proposed Changes

- Fixes #175 
- Removed Tooltip component
- Add CSS styling in `global.css`


AshrafMd-1:
@rithviknishad is this okay

rithviknishad (commented):
Yup! LGTM, do remove the unused `Tooltip` component too

rithviknishad (approved):

PR #232, opened 2024-02-22
Add Git icons to the GithubEvent card title

dgparmar14:
I add icons to the "GithubEvent card title" as per the git hub actions mentioned


app/contributors/dgparmar14.md
rithviknishad, line 1: Discard this file. You can add your markdown to https://github.com/coronasafe/leaderboard-data repo. This repo does not and should not contain any specific data as other organizations would be using this as source code to deploy their data.

contributors
rithviknishad, line 1: This symlink is required for leaderboard to correctly point to the data source

rithviknishad (changes requested):

app/contributors/dgparmar14.md
rithviknishad: Tomorrow morning your markdown profile will be auto-generated by the scrapper workflow. You could also simply edit the existing markdown too then.

contributors
rithviknishad: Still unresolved

rithviknishad (changes requested):

dgparmar14:
Okay i will do it

PR #243, opened 2024-02-25
make all icons responsive for mobile and dekstop view

### LINKED ISSUE #238: Feed: Event title is not responsive
image

dgparmar14:
# Description
Make all icons responsive on the basis of posibble lengths

Fixes #238 

## Type of change
Change only who's length will be creating ui problem

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,16 @@
 {
-  "extends": ["next/core-web-vitals", "plugin:prettier/recommended"],
+  "extends": [
+    "next/core-web-vitals",
+    "plugin:prettier/recommended",
+    "plugin:tailwindcss/recommended"
+  ],
+  "rules": {
+    "tailwindcss/no-custom-classname": "off",
+    "tailwindcss/migration-from-tailwind-2": "error",
+    "tailwindcss/classnames-order": "error",
+    "tailwindcss/enforces-shorthand": "off",
+    "tailwindcss/no-contradicting-classname": "error",
+    "tailwindcss/enforces-negative-arbitrary-values": "off"
+  },
   "ignorePatterns": ["data"]
 }
--- a/.github/workflows/scraper-dry-run.yaml
+++ b/.github/workflows/scraper-dry-run.yaml
@@ -1,0 +1,36 @@
+name: Scraper Dry Run
+
+on:
+  pull_request:
+    paths:
+      - scraper/**
+
+jobs:
+  test-run-github-scraper:
+    name: Test run GitHub Scraper
+    runs-on: ubuntu-latest
+    permissions:
+      issues: read
+      pull-requests: read
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: setup python
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
+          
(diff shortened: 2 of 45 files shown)


rithviknishad (changes requested):
Even though there is enough space for time to be shown in same line, it's going to the next line. Let's keep it in a single line if we have the space

image

dgparmar14:
> Even though there is enough space for time to be shown in same line, it's going to the next line. Let's keep it in a single line if we have the space
> 
> image

Yes,There is a space but I do it because time will be shown below for both Mobile and desktop view .In mobile view there is need to in below line because there is no space.So, I do it same change in the desktop.

rithviknishad:
@dgparmar14 We could simply use `inline-flex` to create an inline-flex container that flows with text. This way what needs to go in new line and what not is automatically handled.

Refer: https://tailwindcss.com/docs/display#inline-flex

dgparmar14:
@rithviknishad In mobile view is it better to below line?

rithviknishad (changes requested):
image

rithviknishad (changes requested):
Why is the whole event title wrapped in a link that opens the related repo?

rithviknishad:
Why? shouldn't it open something related to the event?

dgparmar14:
Okay try to fix it

dgparmar14:
> Why? shouldn't it open something related to the event?

I will change it according to the main leader board repo link will be same as per previous?

rithviknishad:
Yup, revert changes to the link. Not in the scope of this issue/PR

dgparmar14:
> Yup, revert changes to the link. Not in the scope of this issue/PR

Responsiveness is done or any changes needed?

rithviknishad:
image

dgparmar14:
@rithviknishad I think now it is done.

rithviknishad:
> image

This issue is still present

dgparmar14:
Okk, Let me clear for clicking link .
It is an already in main repo code. So whenever link on user name it will navigate to the user's profile

> > image
> 
> This issue is still present

In which screen size it occurs because i try all device on chrome for responsive ness and it is working fine

dgparmar14:
> We can make it responsive by simplifying the component event title to be as shown in the example [here](https://tailwindcss.com/docs/display#inline-flex) regardless of what window size it is.

I already use it but there is also need of flex-wrap so i add it also can you please review and there is still not fixed i will do it

dgparmar14:
Yes, there is still issues I will fix it 😅
when I run it on my device and check responsiveness it's work really fine and in deployment 🙂. can you check code

dgparmar14:
@rithviknishad it's now finally done i checked all possible way can you please review if there is still issue then please telll me.

rithviknishad:
Hey @dgparmar14 could you rebase with the latest main branch and do a `pnpm install` and resolve the conflicts?

rithviknishad (approved):
LGTM, you've made it responsive! Let's merge it as soon as the conflicts are resolved

dgparmar14:
> LGTM, you've made it responsive! Let's merge it as soon as the conflicts are resolved

Sure , I will do it till noon .

Issue #153, opened 2023-12-08
Highlight new contributors that have made their first contribution in the past 1 week in landing page


rithviknishad:
In landing page, add a section to highlight the new contributors based on the condition "made their first contribution in the past 1 week".

khavinshankar:
@rithviknishad can u assign this to me

rhithesh:
Hi I would like to work on this.

rhithesh:
@rithviknishad  please check and let me know if any changes are needed.

androemeda:
@rithviknishad please assign this issue to me

Indrajeety993648:
@rithviknishad  I would like to work on this issue, please assign it to me

syedfardeenjeelani:
@rithviknishad hello i would like to try to work on this issue

syedfardeenjeelani:
hi @rithviknishad  i wanted to  ask should i make a section below our contributors and name it new contributors which will have a ui like our contributors there i will filter the new contributors that have made their first contribution in the past week ?? or do i have to do something different please let me know

rithviknishad:
Let's keep it simple. Let's simply keep a badge beside their name to show that the contributor is a new contributor.

jayvaliya:
hi @rithviknishad can you please assign it to me? 
and also do I have to wait unit issue is assigned to me or should I start working on it directly?

jayvaliya:
![Image](https://github.com/user-attachments/assets/3a239017-8c57-4385-9194-6be9ae533d51)

Hey @rithviknishad 
I have done something like this. Please guide me if you wanted something else.

IMPORTANT QUESTION : Do i need to check the date first pr is raised or the date when first pr merged?

Neha611:
Hey I want to work on this issue @rithviknishad Please assign me this.

rithviknishad:
@jayvaliya missed your comment.

First PR merged sounds good!

JavidSumra:
Hey @rithviknishad can you please assign this issue to me?

PR #264, opened 2024-02-27
adding new Contributors option.

rhithesh:
#153  fixes.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,16 @@
 {
-  "extends": ["next/core-web-vitals", "plugin:prettier/recommended"],
+  "extends": [
+    "next/core-web-vitals",
+    "plugin:prettier/recommended",
+    "plugin:tailwindcss/recommended"
+  ],
+  "rules": {
+    "tailwindcss/no-custom-classname": "off",
+    "tailwindcss/migration-from-tailwind-2": "error",
+    "tailwindcss/classnames-order": "error",
+    "tailwindcss/enforces-shorthand": "off",
+    "tailwindcss/no-contradicting-classname": "error",
+    "tailwindcss/enforces-negative-arbitrary-values": "off"
+  },
   "ignorePatterns": ["data"]
 }
--- a/.github/workflows/scraper-dry-run.yaml
+++ b/.github/workflows/scraper-dry-run.yaml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/setup-python@v3
         with:
           python-version: "3.10"
-          
+
       - run: rm data contributors && mkdir data contributors
 
       - name: Install dependencies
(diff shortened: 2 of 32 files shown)


lib/types.ts
```diff
@@ -39,6 +39,7 @@ export interface Highlights {
   pr_collaborated: number;
   issue_assigned: number;
   issue_opened: number;
+  newContributor?: boolean;
```
rithviknishad, line 42: @rhithesh let's store the `first_activity` date instead of a boolean field as that'd allow us to show "Contributing since <date>" in contributors' profiles, and filtering can be done by doing a comparison of is within the last N days or not against this field.

rithviknishad, line 42: Also, move this field to the "Contributor" type instead of keeping it under "Highlights", since it's specific to a contributor. Highlights are used in multiple other ways where field `newContributor` wouldn't be a right fit.

lib/api.ts
```diff
@@ -142,17 +142,28 @@ export async function getContributorBySlug(file: string, detail = false) {
...
+      HighlightsReducer,
+      HighlightsInitialValue,
+    ) as Highlights;
+    final["newContributor"] = newContributor;
+
+    return final;
   };
 
-  return {
+  let a = {
```
rithviknishad, line 166: Revert unrelated changes

app/api/leaderboard/functions.ts
```diff
@@ -32,17 +32,22 @@ export const getLeaderboardData = async (
...
-    .filter((contributor) => contributor.summary.points)
+    .filter((contributor) => {
+      return contributor.summary.points;
+    })
     .sort((a, b) => {
       if (sortBy === "pr_stale") {
         return b.activityData.pr_stale - a.activityData.pr_stale;
+      } else if (sortBy == "new_contributors") {
+        return b.summary["points"] - a.summary["points"];
       }
```
rithviknishad, line 51: `new_contributors` is a boolean field, how would sort happen exactly? And why is it sorted by points here?

Sort by "date of first activity" instead

app/api/leaderboard/functions.ts
```diff
@@ -32,17 +32,22 @@ export const getLeaderboardData = async (
...
   const data = contributors
     .filter((a) => a.highlights.points)
     .map((contributor) => ({
       ...contributor,
       summary: contributor.summarize(...dateRange),
     }))
-    .filter((contributor) => contributor.summary.points)
+    .filter((contributor) => {
+      return contributor.summary.points;
+    })
```
rithviknishad, line 45: revert unrelated changes

app/api/leaderboard/functions.ts
```diff
@@ -32,17 +32,22 @@ export const getLeaderboardData = async (
   const shouldReverse = !ordering.startsWith("-");
 
   const contributors = await getContributors();
+  // console.log(contributors);
```
rithviknishad, line 35: revert unrelated changes

rithviknishad (changes requested):

PR #280, opened 2024-02-28
update user role field to be single role 

UdaySagar-Git:
Fixes #271

linked pr https://github.com/coronasafe/leaderboard-data/pull/17

PR #286, opened 2024-02-28
Fixes sorting logic

UdaySagar-Git:
- Fixed sorting issue 
- Changed existing logic from using - sign to use ordering search param , this makes code look cleaner


rithviknishad (commented):
Changes from #280 is present here. Was that intentional or is this PR dependent on it? @UdaySagar-Git

UdaySagar-Git:
basically that line caused me typescript errors , since earlier pr have some changed types so had intensionally made a branch on existing branch @rithviknishad 

image

Issue #278, opened 2024-02-28
Change the UI of the Navbar to a more modern and interactive navbar


Lobi29:
![leaderboard](https://github.com/coronasafe/leaderboard/assets/108052802/ee91839a-e125-4238-95d1-a0e5462599a7)

<h3 align="center">I wanted to propose some enhancements to our Navbar that would make it more modern, interactive, and aesthetically pleasing. Here are the key features and improvements I suggest:</h1>

### Contribute Now Option
1. Introduce a prominent "Contribute Now" option within the Navbar, providing users with a clear call-to-action to encourage engagement and support for our platform's goals.

### Decreasing Navbar Height
2. Currently, the Navbar's height appears bulky and may not align with modern design trends. I propose reducing the height to a more modest and visually appealing size, ensuring it integrates seamlessly with the overall website layout.

### Animated Border Middle Div
3. Implement animation effects for the middle div of the Navbar's border to add dynamism and interactivity. This subtle animation will enhance user engagement and create a more polished user interface.

### Transparent and Sticky Navbar:
4. Transform the Navbar to be transparent and sticky, allowing it to remain visible at the top of the page while users scroll through content. This design approach not only offers a sleek and modern appearance but also ensures easy access to navigation options throughout the browsing experience.

HARSHDOIPHODE:
Can anyone assign me this issue?

PR #283, opened 2024-02-28
Adds support for organisation specific theme

dgparmar14:
# Description

This pull request addresses issue #249 by reorganizing the structure and configuration files related to personalized leaderboard colors.

## Changes Made

- Moved the `tailwind.config.js` file from the `leaderboard` repository to the `leaderboard-data` repository.
- Created a new `tailwind.config.js` file in the `leaderboard-data` repository.
- Established a symbolic link (symlink) between the `tailwind.config.js` file in the `leaderboard-data` repository and the `leaderboard` repository.
- Renamed the `load-fixtures.sh` script to `load-org-data.sh` for clarity and consistency.
- Updated the `load-org-data.sh` script to include the logic for fetching the `tailwind.config.js` file from the `leaderboard-data` repository.

## How It Works

After cloning the repository, users can run the command `./load-org-data.sh`, which fetches the data and `tailwind.config.js` file from the `leaderboard-data` repository, enabling personalized leaderboard colors.

This restructuring improves organization and maintainability while ensuring consistent behavior across environments.

Fixes: #249

PR #287, opened 2024-02-29
updating UI of Navbar

Lobi29:
<h1 align="center">ISSUE #278</h1>

### Feature Added

- "Contribute Now" option link
- Navbar made sticky with backdrop blur effect
- Direct links to leaderboard, contributors, feed, and projects
- Full Responsive

### Mobile View:

Screenshot from 2024-02-29 10-08-13

Screenshot from 2024-02-29 10-08-05

### Desktop View

Screenshot from 2024-02-29 10-07-11
Screenshot from 2024-02-29 10-06-58


Lobi29:
### changing the gradient to a more asthetic and minimal one

Untitled drawing

- if it's all right, then please merge it 😊

components/navbar/ContributeButton.tsx
```diff
@@ -0,0 +1,21 @@
+"use client";
```
rithviknishad, line 1: Why is this a `use client` component?

components/navbar/ContributeButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Link from "next/link";
+import { AiFillGithub } from "react-icons/ai";
+
+const ContributeButton = () => {
+  return (
+    <Link
+      href="https://github.com/coronasafe/leaderboard/issues"
```
rithviknishad, line 8: Leaderboard would be deployed by different organizations, it'd be great to open the organization as the viewer might be looking to contribute to the organization.

Organization-specific data is not kept in this repo.
Use environment variables.

app/globals.css
```diff
@@ -136,3 +136,9 @@ html.dark {
   transform: translateX(0px);
   transform: translateY(0px);
 }
+
+.text-gradient {
+  background: linear-gradient(45deg, #6366f1, #1d4ed8);
```
rithviknishad, line 141: What if the organization deploying the leaderboard has a different colour theme? #283

components/navbar/Logo.tsx
```diff
@@ -0,0 +1,18 @@
+"use client";
```
rithviknishad, line 1: Why use client again?

```suggestion
```

components/navbar/MenuItems.tsx
```diff
@@ -0,0 +1,31 @@
+"use client";
```
rithviknishad, line 1: Why use client?

rithviknishad (changes requested):

components/navbar/ContributeButton.tsx
Lobi29: sorry for this silly mistake, I am removing this line from this component

components/navbar/ContributeButton.tsx
Lobi29: ok, I will update it with .env varible

app/globals.css
Lobi29: okay, I will use the colour theme from the tailwind.config.ts file

components/navbar/Logo.tsx
Lobi29: again sorry

components/navbar/MenuItems.tsx
Lobi29: again again sorry 😊

Lobi29:
<h1 align="center">Updates</h1>

- Added `.env` variable to the contribution link's href
- Color theme applied, with emphasis on the primary theme
- 'use client' feature removed from redundant components

rithviknishad (changes requested):
@Lobi29 

1. The gradient still feels like the text is not very readable. More contrast between the background is needed. Let's keep it as minimal and aesthetic as possible.

image


2. Let's keep the old sidebar in the mobile view itself. This is because in future we may have more links to be shown.

image

3. The theme switcher is getting more attention than required due to the strong contrast from the background. Let's keep the colors of the theme switcher like how it was before.

image

4. The navbar looked good before. 
image

Lobi29:
Got it 😁👍,

Lobi29:
color is updated

rithviknishad (changes requested):
@Lobi29 

I don't think this sidebar would look good on mobile. Let's have the old sidebar that takes the full height of the screen itself and show the theme switcher inside the sidebar in mobile view. Let's not do blur for the slide over part of the sidebar as it'd make the text unreadable. (Fine with navbar).

image



Also, you've not made the text lighter yet to bring in more contrast from the background.
It is still not very readable.

image

image

rithviknishad (changes requested):
Also use the default font configured (that is font-mono) across the leaderboard. Having font-inter in just some places of leaderboard doesn't make it look good.

Lobi29:
Untitled drawing (4)

- Drawing from your previous advice, I've refined the sidebar and subtly adjusted the text's hue, enhancing its readability and aesthetic appeal

rithviknishad (commented):
LGTM, just one last suggestion

image

Lobi29:
sure, why not 😊

Lobi29:
Hey @rithviknishad , "Contribute Now" text has been removed, leaving only the GitHub button visible.😊

rithviknishad (approved):

PR #309, opened 2024-03-02
Add rank of contributor on the weeklybasis in profile page

### LINKED ISSUE #295: Show rank of contributor inside the contributor profile page
image

image

dgparmar14:
# Description
To show a rank of contributor in profile page. Function created in `api.ts` to fetch contributor rank in `infocard.tsx`

Fixes #295 

## Type of change
`/lib/api.ts`
`/components/contributors/InforCrad`
`app/contributors/[slug]/page.tsx`

### DIFF
--- a/app/contributors/[slug]/page.tsx
+++ b/app/contributors/[slug]/page.tsx
@@ -14,13 +14,14 @@ import InfoCard from "../../../components/contributors/InfoCard";
 import React, { Suspense } from "react";
 import clsx from "clsx";
 import { Contributor } from "@/lib/types";
-import { formatDuration } from "@/lib/utils";
+import { formatDuration, parseDateRangeSearchParam } from "@/lib/utils";
 import Markdown from "@/components/Markdown";
 import { FiAlertTriangle } from "react-icons/fi";
 import { TbGitPullRequest } from "react-icons/tb";
 import RelativeTime from "@/components/RelativeTime";
 import Link from "next/link";
 import { env } from "@/env.mjs";
+import { getLeaderboardData } from "@/app/api/leaderboard/functions";
 
 export async function generateStaticParams() {
   const slugs = await getContributorsSlugs();
@@ -39,6 +40,18 @@ export default async function Contributor({ params }: Params) {
   const { slug } = params;
   const contributor = await getContributorBySlug(slug, true);
 
+  const leaderboardData = await getLeaderboardData(
+    parseDateRangeSearchParam(),
+    "points",
+    "desc",
+    [],
+  );
+  const userPosition = leaderboardData.findIndex(
+    (data) =>
+      data.user.social.github === contributor.github &&
+      contributor.weekSummary.points !== 0,
+  );
+
   return (
     <div className="min-h-screen bg-background">
       {/* <Header /> */}
@@ -48,9 +61,9 @@ export default async function Contributor({ params }: Params) {
         </h1>
       </div>
       <section className="bg-secondary-200 px-4 py-8 dark:bg-secondary-800">
-        <div className="mx-auto flex max-w-6xl flex-col md:flex-row lg:space-x-12 ">
-          <div className="md:w-2/3">
-            <InfoCard contributor={contributor} showRank />
+        <div className="mx-auto flex max-w-6xl flex-col md:flex-row lg:gap-16">
+          <div className="min-w-max">
+            <InfoCard contributor={contributor} rank={userPosition + 1} />
           </div>
           <div className="mt-6 flex w-full gap-2 overflow-x-auto md:mt-0 md:grid md:grid-cols-7 md:overflow-x-visible">
             {[
--- a/app/contributors/[slug]/page.tsx
+++ b/app/contributors/[slug]/page.tsx
@@ -61,12 +61,12 @@ export default async function Contributor({ params }: Params) {
           Personal Learning Dashboard (Beta)
         </h1>
       </div>
-      <section className="bg-secondary-200 px-4 py-8 dark:bg-secondary-800">
-        <div className="mx-auto flex max-w-6xl flex-col md:flex-row lg:gap-16">
-          <div className="min-w-max">
+      <section className="bg-secondary-200 px-4 py-6 dark:bg-secondary-800">
+        <div className=" mx-auto flex max-w-6xl flex-col gap-2 md:flex-row lg:gap-16">
+          <div className="mx-auto my-auto min-w-max md:w-2/3">
             <InfoCard contributor={contributor} rank={rank} />
           </div>
-          <div className="mx-1/7 flex flex-wrap justify-center">
+          <div className="mb-2 flex flex-wrap justify-center lg:grid lg:w-full lg:grid-cols-7 lg:gap-2">
             {[
               professionalSelfSkills,
               professionalTeamSkills,
@@ -75,7 +75,7 @@ export default async function Contributor({ params }: Params) {
             ].map((attributeGroup) => {
               return attributeGroup.map((skill) => (
                 <div
-                  className="mx-2 mt-3 flex shrink-0 items-center justify-center rounded-lg lg:mx-5"
+                  className="mx-2 mt-3 shrink-0 items-center justify-center rounded-lg lg:mx-5"
                   key={skill.key}
                 >
                   <BadgeIcons
--- a/app/contributors/[slug]/page.tsx
+++ b/app/contributors/[slug]/page.tsx
@@ -48,9 +48,9 @@ export default async function Contributor({ params }: Params) {
         </h1>
       </div>
       <section className="bg-secondary-200 px-4 py-8 dark:bg-secondary-800">
-        <div className="mx-auto flex max-w-6xl flex-col md:flex-row ">
+        <div className="mx-auto flex max-w-6xl flex-col md:flex-row lg:space-x-12 ">
           <div className="md:w-2/3">
-            <InfoCard contributor={contributor} />
+            <InfoCard contributor={contributor} showRank />
           </div>
           <div className="mt-6 flex w-full gap-2 overflow-x-auto md:mt-0 md:grid md:grid-cols-7 md:overflow-x-visible">
             {[
--- a/app/contributors/[slug]/page.tsx
+++ b/app/contributors/[slug]/page.tsx
@@ -66,7 +66,7 @@ export default async function Contributor({ params }: Params) {
           <div className="min-w-max">
             <InfoCard contributor={contributor} rank={rank} />
           </div>
-          <div className="mt-6 flex w-full gap-2 overflow-x-auto md:mt-0 md:grid md:grid-cols-7 md:overflow-x-visible">
+          <div className="mx-1/7 flex flex-wrap justify-center">
             {[
               professionalSelfSkills,
               professionalTeamSkills,
@@ -75,7 +75,7 @@ export default async function Contributor({ params }: Params) {
             ].map((attributeGroup) => {
               return attributeGroup.map((skill) => (
                 <div
-                  className="flex shrink-0 items-center justify-center rounded-lg"
+                  className="mx-2 mt-3 flex shrink-0 items-center justify-center rounded-lg lg:mx-5"
                   key={skill.key}
                 >
                   <BadgeIcons
--- a/app/contributors/[slug]/page.tsx
+++ b/app/contributors/[slug]/page.tsx
@@ -46,11 +46,12 @@ export default async function Contributor({ params }: Params) {
     "desc",
     [],
   );
-  const userPosition = leaderboardData.findIndex(
-    (data) =>
-      data.user.social.github === contributor.github &&
-      contributor.weekSummary.points !== 0,
-  );
+  const rank =
+    leaderboardData.findIndex(
+      (data) =>
+        data.user.social.github === contributor.github &&
+        contributor.weekSummary.points !== 0,
+    ) + 1 || null;
 
   return (
     <div className="min-h-screen bg-background">
@@ -63,7 +64,7 @@ export default async function Contributor({ params }: Params) {
       <section className="bg-secondary-200 px-4 py-8 dark:bg-secondary-800">
         <div className="mx-auto flex max-w-6xl flex-col md:flex-row lg:gap-16">
           <div className="min-w-max">
-            <InfoCard contributor={contributor} rank={userPosition + 1} />
+            <InfoCard contributor={contributor} rank={rank} />
           </div>
           <div className="mt-6 flex w-full gap-2 overflow-x-auto md:mt-0 md:grid md:grid-cols-7 md:overflow-x-visible">
             {[
--- a/app/contributors/[slug]/page.tsx
+++ b/app/contributors/[slug]/page.tsx
@@ -352,7 +352,7 @@ export default async function Contributor({ params }: Params) {
                   Last contribution{" "}
                   {contributor.activityData.last_updated ? (
                     <RelativeTime
-                      time={contributor.activityData.last_updated * 1e3}
+                      time={contributor.activityData.last_updated}
                     />
                   ) : (
                     "unknown"
--- a/components/contributors/InfoCard.tsx
+++ b/components/contributors/InfoCard.tsx
@@ -5,16 +5,28 @@ import Image from "next/image";
 import Link from "next/link";
 import { BsSlack } from "react-icons/bs";
 import { env } from "@/env.mjs";
+import { getLeaderboardData } from "@/app/api/leaderboard/functions";
+import { parseDateRangeSearchParam } from "@/lib/utils";
 
-export default function InfoCard({
+export default async function InfoCard({
   contributor,
-  minimal = true,
+  showRank,
   isClickable = false,
 }: {
   contributor: Contributor;
-  minimal?: boolean;
+  showRank?: boolean;
   isClickable?: boolean;
 }) {
+  const leaderboardData = await getLeaderboardData(
+    parseDateRangeSearchParam(),
+    "points",
+    "desc",
+    [],
+  );
+  const userPosition = leaderboardData.findIndex(
+    (data) => data.user.social.github === contributor.github,
+  );
+
   return (
     <div
       className={clsx(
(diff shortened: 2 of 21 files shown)


dgparmar14:
@rithviknishad Hey, Rithvik can you please review it?

app/contributors/[slug]/page.tsx
```diff
@@ -50,7 +49,7 @@ export default async function Contributor({ params }: Params) {
       <section className="bg-gray-200 px-4 py-8 dark:bg-gray-800">
         <div className="mx-auto flex max-w-6xl flex-col md:flex-row ">
           <div className="md:w-2/3">
-            <InfoCard contributor={contributor} />
+            <InfoCard contributor={contributor} isRank={true} />
```
Comment (rithviknishad) [inline app/contributors/[slug]/page.tsx:52]: Rely on the `minimal` prop instead

```suggestion
            <InfoCard contributor={contributor} />
```

lib/api.ts
```diff
@@ -300,3 +300,12 @@ const getLastWeekHighlights = (calendarData: Highlights[]) => {
 
   return highlights;
 };
+
+export const getContributorRank = async (name: string) => {
```
rithviknishad, line 304: I'd say add a prop in the `weekSummary` itself specifying the `rank` as a field instead of a new function

rithviknishad (changes requested):

lib/api.ts
dgparmar14: I try that, sure I will do it again

app/contributors/[slug]/page.tsx
Comment (dgparmar14) [reply app/contributors/[slug]/page.tsx:52]: we are using githubevent component in home page also so I use props to check rank is shown or not

Comment (rithviknishad) [reply app/contributors/[slug]/page.tsx:52]: GitHub event has nothing to do with InfoCard IIRC

Comment (dgparmar14) [reply app/contributors/[slug]/page.tsx:52]: Sorry, I mean InfoCard we are showing in the home page also, 
image

dgparmar14:
I do not remove isRank props because it indicates to show the ranking in the profile page and if we don't add it the ranking will also show in the home page info cards does it right?
One more thing is that I fetched leaderboard data and use index as ranking only .And for contributor data i don't chnage previous logic.

app/contributors/[slug]/page.tsx
Comment (rithviknishad) [reply app/contributors/[slug]/page.tsx:52]: Yea, `minimal` is the prop that differentiates from whats shown in home page and whats shown in contributor profile.

If minimal is true, skip showing rank

components/contributors/InfoCard.tsx
```diff
@@ -5,16 +5,31 @@ import Image from "next/image";
...
+  const leaderboardData = await getLeaderboardData(
+    parseDateRangeSearchParam(),
+    "points",
+    "desc",
+    [],
+  );
+  const userPosition = leaderboardData.findIndex(
+    (data) => data.user.social.github === contributor.github,
+  );
+  const hideBadges = userPosition === -1;
```
rithviknishad, line 31: It's fine to inline this in the JSX itself. Also even though you are naming it as `hideBadges` this variable seems to have nothing to do with the badges.

rithviknishad (changes requested):

dgparmar14:
@rithviknishad  If use minimal to show rank it will create logical conflicts in info card page beacause the default value of minimal is true

rithviknishad:
> @rithviknishad If use minimal to show rank it will create logical conflicts in info card page beacause the default value of minimal is true

We also don't want to show the rank in `minmal=true` situations either right?

rithviknishad:
Oh yea, makes sense. If you think the `minimal` is not used anywhere, let's remove it and keep `showRank` prop instead

dgparmar14:
> Oh yea, makes sense. If you think the `minimal` is not used anywhere, let's remove it and keep `showRank` prop instead

So, I need to change my previous commit?

dgparmar14:
@rithviknishad Can you please review it

components/contributors/InfoCard.tsx
```diff
@@ -5,16 +5,30 @@ import Image from "next/image";
...
+export default async function InfoCard({
   contributor,
-  minimal = true,
+  isRank,
+  // minimal = true,
   isClickable = false,
 }: {
   contributor: Contributor;
-  minimal?: boolean;
+  isRank?: boolean;
```
rithviknishad, line 18: Repeating... rename it to `showRank` as `isRank` isn't an appropriate name for what this prop does.

```suggestion
  showRank?: boolean;
```

components/contributors/InfoCard.tsx
```diff
@@ -94,7 +107,7 @@ export default function InfoCard({
                         d="M12 2C6.477 2 2 6.463 2 11.97c0 4.404 2.865 8.14 6.839 9.458.5.092.682-.216.682-.48 0-.236-.008-.864-.013-1.695-2.782.602-3.369-1.337-3.369-1.337-.454-1.151-1.11-1.458-1.11-1.458-.908-.618.069-.606.069-.606 1.003.07 1.531 1.027 1.531 1.027.892 1.524 2.341 1.084 2.91.828.092-.643.35-1.083.636-1.332-2.22-.251-4.555-1.107-4.555-4.927 0-1.088.39-1.979 1.029-2.675-.103-.252-.446-1.266.098-2.638 0 0 .84-.268 2.75 1.022A9.606 9.606 0 0112 6.82c.85.004 1.705.114 2.504.336 1.909-1.29 2.747-1.022 2.747-1.022.546 1.372.202 2.386.1 2.638.64.696 1.028 1.587 1.028 2.675 0 3.83-2.339 4.673-4.566 4.92.359.307.678.915.678 1.846 0 1.332-.012 2.407-.012 2.734 0 .267.18.577.688.48C19.137 20.107 22 16.373 22 11.969 22 6.463 17.522 2 12 2z"
                       ></path>
                     </svg>
-                    {!minimal && `@${contributor.github}`}
+                    {/* {!minimal && `@${contributor.github}`} */}
```
rithviknishad, line 110: Let's remove these instead of commenting it out.

components/contributors/InfoCard.tsx
```diff
@@ -5,16 +5,30 @@ import Image from "next/image";
...
 import { env } from "@/env.mjs";
+import { getLeaderboardData } from "@/app/api/leaderboard/functions";
+import { parseDateRangeSearchParam } from "@/lib/utils";
 
-export default function InfoCard({
+export default async function InfoCard({
   contributor,
-  minimal = true,
+  isRank,
+  // minimal = true,
```
rithviknishad, line 14:
```suggestion
```

components/contributors/InfoCard.tsx
```diff
@@ -5,16 +5,30 @@ import Image from "next/image";
...
   isClickable = false,
 }: {
   contributor: Contributor;
-  minimal?: boolean;
+  isRank?: boolean;
+  // minimal?: boolean;
```
rithviknishad, line 19:
```suggestion
```

components/contributors/InfoCard.tsx
```diff
@@ -24,31 +38,35 @@ export default function InfoCard({
       )}
       role="listitem"
     >
-      <div className="flex shrink-0 items-center space-x-4 md:space-y-6 xl:space-y-1 ">
+      <div className="flex shrink-0 items-center space-x-2 md:space-y-6 xl:space-y-1 ">
+        {isRank && (
+          <div
+            className={`flex h-20 w-20 items-center text-5xl font-[500] text-gray-500 dark:text-gray-400 lg:text-6xl`}
```
rithviknishad, line 44:
```suggestion
            className="flex h-20 w-20 items-center text-5xl font-bold text-gray-500 dark:text-gray-400 lg:text-6xl tracking-wider"
```

rithviknishad (changes requested):

dgparmar14:
@rithviknishad  All your suggested changes are done please review it

components/contributors/InfoCard.tsx
```diff
@@ -68,12 +82,7 @@ export default function InfoCard({
             </p>
           </div>
 
-          <ul
-            role="list"
-            className={
-              minimal ? "mt-4 flex items-center space-x-4" : "space-y-2"
-            }
-          >
+          <ul role="list" className={"mt-4 flex items-center space-x-4"}>
```
rithviknishad, line 85:
```suggestion
          <ul role="list" className="mt-4 flex items-center space-x-4">
```

app/contributors/[slug]/page.tsx
```diff
@@ -50,7 +49,7 @@ export default async function Contributor({ params }: Params) {
       <section className="bg-gray-200 px-4 py-8 dark:bg-gray-800">
         <div className="mx-auto flex max-w-6xl flex-col md:flex-row ">
           <div className="md:w-2/3">
-            <InfoCard contributor={contributor} />
+            <InfoCard contributor={contributor} showRank={true} />
```
Comment (rithviknishad) [inline app/contributors/[slug]/page.tsx:52]:
```suggestion
            <InfoCard contributor={contributor} showRank />
```

rithviknishad (approved):

rithviknishad (changes requested):
Improper spacing between elements. Needs more whitespace. The rank and profile picture area looks more cramped.

**Before**

image

**After**

image

rithviknishad (changes requested):
The spacing issue is still not solved. Kindly test before requesting for review

rithviknishad:
Yup

dgparmar14:
> resolve conflicts

Have you resolve conflicts? I don't know why my closed

dgparmar14:
@rithviknishad Can you verify it? or still there is conflicts  i will  solve it

components/contributors/InfoCard.tsx
```diff
@@ -63,25 +79,20 @@ export default function InfoCard({
                 {contributor.name}
               </h3>
             </Link>
-            <p className="text-sm text-secondary-400 md:text-base">
+            <p className="text-sm text-gray-400 md:text-base">
```
rithviknishad, line 82: Ensure conflicts are resolved properly. We are no longer using `gray` shade directly as secondary colors are configurable. Use `secondary` shade instead.

components/contributors/InfoCard.tsx
```diff
@@ -63,25 +79,20 @@ export default function InfoCard({
...
+          <ul role="list" className="mt-4 flex items-center space-x-4">
             {contributor.github && (
               <li role="listitem">
                 <Link
                   href={`https://github.com/${contributor.github}`}
                   target="_blank"
                   rel="noreferrer"
                 >
-                  <span className="flex items-center text-secondary-500 hover:text-primary-300">
+                  <span className="flex items-center text-gray-500 hover:text-primary-300">
```
rithviknishad, line 95: here too

components/contributors/InfoCard.tsx
```diff
@@ -106,7 +116,7 @@ export default function InfoCard({
                   target="_blank"
                   rel="noreferrer"
                 >
-                  <div className="flex items-center text-secondary-500 hover:text-primary-300">
+                  <div className="flex items-center text-gray-500 hover:text-primary-300">
```
rithviknishad, line 119: here...

components/contributors/InfoCard.tsx
```diff
@@ -128,7 +137,7 @@ export default function InfoCard({
...
```
rithviknishad, line 140: here..

components/contributors/InfoCard.tsx
```diff
@@ -154,10 +162,9 @@ export default function InfoCard({
...
```
rithviknishad, line 165: why was all secondary changed to gray even with places without conflicts?

dgparmar14: I think i make some mistakes so, I will resolve ASAP

rithviknishad (changes requested):
- [ ] 2 digit ranks has an overlapping issue

image

- [ ] If I visit a profile where they are not part of the current weeks leaderboard as they've not made any contributions, their ranks are shown as `0` which is incorrect. Instead we may not show any rank and keep it like how we had before.

image

- [ ] Change the ranks color to be same as the color used for first, second and third's shadow/border in image and for rest let's use the primary shade.

components/contributors/InfoCard.tsx
```diff
@@ -5,16 +5,28 @@ import Image from "next/image";
...
 }) {
+  const leaderboardData = await getLeaderboardData(
+    parseDateRangeSearchParam(),
+    "points",
+    "desc",
+    [],
+  );
+  const userPosition = leaderboardData.findIndex(
+    (data) => data.user.social.github === contributor.github,
+  );
```
rithviknishad, line 28: This compute is performed for each and every render of `InfoCard` component. For example, in the home page, we are showing this card N number of times, but we are not using the rank computed from this.

This could make it slow when we have larger amount of data.

Instead move the compute to the parent component and pass the rank down to this component only in places where it is required.

rithviknishad (changes requested):

components/contributors/InfoCard.tsx
dgparmar14: If we fetch in the `contributor's` page which is the parent of `Infocard`. On that page, we have the `getContributorBySlug()` function and In the same page if we fetch from leaderboard data using `getLeaderboardData()` it creates partial deadlock. Right?

So, To resolve the performance issue we can apply conditional fetching if the show rank is true only then fetch data from the leaderboard. We can do this using the `useEffect` or `usePathname` as well.

rithviknishad: The deadlock wouldn't happen in this case since both are independently executed.

pnpm-lock.yaml
rithviknishad, line 1: Discard the changes in lock file

components/contributors/InfoCard.tsx
```diff
@@ -6,13 +6,13 @@ import Link from "next/link";
 import { BsSlack } from "react-icons/bs";
 import { env } from "@/env.mjs";
 
-export default function InfoCard({
+export default async function InfoCard({
   contributor,
-  minimal = true,
+  rank = 0,
```
rithviknishad, line 11: set it to `null` by default In situations where `rank` is null, we can hide it.

```suggestion
  rank,
```

components/contributors/InfoCard.tsx
```diff
@@ -24,31 +24,35 @@ export default function InfoCard({
...
-          href={isClickable ? `/contributors/${contributor.github}` : `#`}
-          className=""
-        >
-          <div
-            className={`z-10 flex h-28 w-28 shrink-0 items-center rounded-full md:h-32 md:w-32 md:p-1 ${
-              isClickable && `cursor-pointer`
-            }`}
+      <div className="flex shrink-0 items-center space-x-2 md:space-y-6 xl:space-y-1 ">
+        <div className="flex">
+          {rank > 0 && (
```
rithviknishad, line 29:
```suggestion
          {!!rank && (
```

app/contributors/[slug]/page.tsx
```diff
@@ -39,6 +40,18 @@ export default async function Contributor({ params }: Params) {
...
+    parseDateRangeSearchParam(),
+    "points",
+    "desc",
+    [],
+  );
+  const userPosition = leaderboardData.findIndex(
+    (data) =>
+      data.user.social.github === contributor.github &&
+      contributor.weekSummary.points !== 0,
+  );
```
Comment (rithviknishad) [inline app/contributors/[slug]/page.tsx:53]:
```suggestion
  const rank = leaderboardData.findIndex(
    (data) =>
      data.user.social.github === contributor.github &&
      contributor.weekSummary.points !== 0,
  ) + 1 || null;
```

app/contributors/[slug]/page.tsx
```diff
@@ -48,9 +61,9 @@ export default async function Contributor({ params }: Params) {
         </h1>
       </div>
       <section className="bg-secondary-200 px-4 py-8 dark:bg-secondary-800">
-        <div className="mx-auto flex max-w-6xl flex-col md:flex-row ">
-          <div className="md:w-2/3">
-            <InfoCard contributor={contributor} />
+        <div className="mx-auto flex max-w-6xl flex-col md:flex-row lg:gap-16">
+          <div className="min-w-max">
+            <InfoCard contributor={contributor} rank={userPosition + 1} />
```
Comment (rithviknishad) [inline app/contributors/[slug]/page.tsx:66]:
```suggestion
            <InfoCard contributor={contributor} rank={rank} />
```

rithviknishad (changes requested):

components/contributors/InfoCard.tsx
```diff
@@ -24,31 +24,37 @@ export default function InfoCard({
...
-          `}
-        >
+            <div
+              className={`dark:border-1 z-10 mr-2 shrink-0 rounded-full border-2 border-current ${
+                rank != null
+                  ? ["text-yellow-600", "text-stone-600", "text-amber-700"][
+                      rank - 1
+                    ] ?? "text-purple-600"
+                  : "text-purple-600"
+              } rounded-full ${rank != null && rank > 0 && rank <= 3 && "animate-circular-shadow"}`}
```
rithviknishad, line 45: Create a reusable function that returns the desired classNames based on rank and re-use that for both in the leaderboard contributor image and here. This way it's guaranteed to always have same logic.

rithviknishad (changes requested):

components/contributors/InfoCard.tsx
dgparmar14: image
Should I keep this function in the `InfoCard.jsx` or move it in `utils.js` ?

rithviknishad: utils

lib/utils.ts
```diff
@@ -93,3 +94,11 @@ export const navLinks = [
...
   return format(date, "MMM dd, yyyy");
 };
+
+export function contributorRankClasses(rank: number | null): string {
+  return clsx(
+    rank != null && rank > 0 && rank <= 3
+      ? `${["text-yellow-600", "text-stone-600", "text-amber-700"][rank - 1]} animate-circular-shadow`
+      : "text-purple-600",
+  );
+}
```
rithviknishad, line 104: This code code be made more readable.

```suggestion
export function contributorRankClasses(rank: number | null): string {
  if (!rank || rank > 3) {
    return "text-purple-500";
  }
  
  const rankColor = ["text-yellow-600", "text-stone-600", "text-amber-700"][rank - 1];
  return `${rankColor} animate-circular-shadow`;
}
```

rithviknishad (approved):

lib/utils.ts
dgparmar14: Yup, It's looks more readable and easy to understand, Thanks for suggestion

rithviknishad: Will merge once this is pushed

dgparmar14: Finally merge it after a week 🫡

lib/utils.ts
```diff
@@ -1,3 +1,4 @@
+import clsx from "clsx";
```
rithviknishad, line 1: Remove unused import
```suggestion
```

rithviknishad (approved):

dgparmar14:
It's not showing `animated-circular-shadow`.

dgparmar14:
I think it takes time to call a function and if I put the classes without function in div element it work's fine
should i revert it or what is your suggestion?

components/contributors/LeaderboardCard.tsx
```diff
@@ -48,11 +49,7 @@ export default function LeaderboardCard({
           <div className="flex w-full">
             <div className="flex min-w-0 flex-1 items-center">
               <div
-                className={`dark:border-1 shrink-0 rounded-full border-2 border-current ${
-                  ["text-yellow-600", "text-stone-600", "text-amber-700"][
-                    position
-                  ] ?? "text-purple-600"
-                } rounded-full ${position < 3 && "animate-circular-shadow"}`}
+                className={`dark:border-1 shrink-0 rounded-full border-2 border-current ${contributorRankClasses(userPosition)}`}
```
rithviknishad, line 52:
```suggestion
                className={`dark:border-1 shrink-0 rounded-full border-2 border-current ${contributorRankClasses(position + 1)}`}
```

rithviknishad (changes requested):
Inspect the tailwind config contents

rithviknishad:
https://tailwindcss.com/docs/content-configuration#configuring-source-paths

rithviknishad (approved):

rithviknishad (changes requested):
Content is overflowing in mobile view

dgparmar14:
when you see responsiveness, 
Whole page is not responsive, 
Can we create separate issue for this?

rithviknishad:
The content is not overflowing in current production. The responsiveness below this component can be solved as a separate issue, but this overflow issue needs to be tackled in this PR as this PR introduces that issue.

image

dgparmar14:
Okk, What should I do ?
Make it flex-wrap I mean in New line or
Make it small size of text?

rithviknishad:
We can just put the rank trailing the name with a slightly faded colour in the same font size as the name.


image

dgparmar14:
image

Done okk?

rithviknishad:
Yup

rithviknishad (approved):

PR #310, opened 2024-03-03
Changes Default Values of Roles for Initial Page Load on Leaderboard Page 

### LINKED ISSUE #289: Change Default Values of Roles for Initial Page Load on Leaderboard Page
Currently, the initial page load displays data for all roles. However, to enhance the user experience, it is proposed to show data specifically for interns and contributors by default.

UdaySagar-Git:
fixes #289

### DIFF
--- a/app/leaderboard/_components/LeaderboardWrapper.tsx
+++ b/app/leaderboard/_components/LeaderboardWrapper.tsx
@@ -2,7 +2,7 @@ import { getLeaderboardData } from "../../api/leaderboard/functions";
 import { parseDateRangeSearchParam } from "@/lib/utils";
 import Leaderboard from "./Leaderboard";
 import { LeaderboardPageProps } from "../page";
-import { deaultRoles } from "@/lib/const";
+import { env } from "@/env.mjs";
 
 export default async function LeaderboardWrapper({
   searchParams,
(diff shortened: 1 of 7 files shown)


UdaySagar-Git:
@rithviknishad should i update home page leader board page with default roles or 
let it be for all roles

app/leaderboard/_components/LeaderboardWrapper.tsx
```diff
@@ -16,7 +17,7 @@ export default async function LeaderboardWrapper({
       | "intern"
       | "operations"
       | "contributor"
-    )[]) ?? [],
+    )[]) ?? deaultRoles.map((i) => i.value),
```
rithviknishad, line 20: Let's keep this configurable as different organizations require different behaviour. Move this to the environment: `NEXT_PUBLIC_LEADERBOARD_DEFAULT_ROLES=comma,seperated,values`

lib/const.ts
```diff
@@ -0,0 +1,4 @@
+export const deaultRoles = [
+  { value: "intern", text: "Intern" },
+  { value: "contributor", text: "Contributor" },
+];
```
rithviknishad, line 4: Not required once moved to env

rithviknishad (changes requested):
Also update the home page

UdaySagar-Git:
@rithviknishad can you please review

rithviknishad (approved):

PR #314, opened 2024-03-03
fixes-client-side-exception

### LINKED ISSUE #292: Client side exception in feed page
### Description :
I noticed that if we go on feed page and go down and click on more 3 times it will show a client-side expectations occur. 
image-1

dgparmar14:
# Description

Fixes #292 .

This pull request addresses the client-side exception occurring in the feed page by transitioning it from client-side rendering to `server-side rendering (SSR)`. The exception is triggered when attempting to fetch data within the feed page, as reaching the GitHub rate limit restricts further requests.

To mitigate this issue, we've implemented SSR for the feed page. Data is fetched from the `github.server-action` file using `octokit.js`. This approach allows us to authenticate requests using a personal token and overcome the rate limit imposed by the GitHub REST API, which restricts to `5000` requests per hour when using GitHub personal token.

In the SSR implementation, data is initially fetched with a **per_.page limit of** `1000`. Subsequently, it's revalidated every **10 minutes** by Next.js, ensuring the feed page remains up-to-date while adhering to GitHub's rate limit policy.

## Type of change
 Enhancement: Transitioning from client-side to server-side rendering for the feed page.
 Fetching data using octokit.js in Feed page.


lib/github.server-action.ts
```diff
@@ -0,0 +1,55 @@
+import { IGitHubEvent } from "@/lib/gh_events";
+import { Octokit } from "octokit";
+
+const GITHUB_PAT: string = process.env.GITHUB_PAT || "";
```
rithviknishad, line 4: use `getGitHubAccessToken` method instead as it'll handle the cases where there is no GITHUB_PAT configured gracefully during development

lib/github.server-action.ts
```diff
@@ -0,0 +1,55 @@
...
+    "DeleteEvent",
+    "IssueCommentEvent",
+  ] as IGitHubEvent["type"][];
+
+  return !blacklist.includes(event.type);
+};
+
+export const exludeBotEvents = (event: IGitHubEvent): boolean => {
+  return !event.actor.login.includes("bot");
+};
```
rithviknishad, line 55: Let's keep all these inside the `page.tsx` of the feed itself. It's not going to be re-used anywhere else.

rithviknishad (changes requested):

lib/github.server-action.ts
```diff
@@ -0,0 +1,55 @@
...
+  "ForkEvent",
+  "ReleaseEvent",
+  "IssueCommentEvent",
+];
+
+export async function fetchFilteredGitHubEvents() {
+  try {
+    const response = await octokit.request("GET /orgs/{org}/events", {
+      org: GITHUB_ORG,
+      per_page: 800,
```
rithviknishad, line 26:
```suggestion
      per_page: 1000,
```

lib/github.server-action.ts
```diff
@@ -0,0 +1,55 @@
...
+  "PullRequestReviewEvent",
+  "PushEvent",
+  "ForkEvent",
+  "ReleaseEvent",
+  "IssueCommentEvent",
+];
+
+export async function fetchFilteredGitHubEvents() {
+  try {
+    const response = await octokit.request("GET /orgs/{org}/events", {
```
rithviknishad, line 24: Do a `.paginate` instead as that'll fetch all the available pages. You can filter the items inside the callback itself to reduce maemory usage

lib/github.server-action.ts
dgparmar14: Sure , I will do it I thought that it will going to reuse at sometime so.

rithviknishad (changes requested):

rithviknishad:
@dgparmar14 repeating again kindly link the related issue using closing keywords https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

app/feed/page.tsx
```diff
@@ -1,94 +1,54 @@
...
-  const allEvents = ([] as IGitHubEvent[]).concat(...Object.values(events));
-
+export default async function FeedPage({ searchParams }: Props) {
+  const events = await octokit.paginate(
+    "GET /orgs/{org}/events",
+    {
+      org: GITHUB_ORG,
+      per_page: 1000,
+    },
+    (response: any) => {
```
rithviknishad, line 28: OctoKit provides typesafe responses. Avoid using any
```suggestion
    (response) => {
```

rithviknishad (changes requested):

dgparmar14:
@rithviknishad Rithvik can you please review it?

dgparmar14:
@rithviknishad Can you please merge and close the issue

app/feed/page.tsx
```diff
@@ -1,94 +1,54 @@
...
-import { scrollTo } from "@/lib/utils";
 import LoadingText from "@/components/LoadingText";
-import { env } from "@/env.mjs";
+import { IGitHubEvent } from "@/lib/gh_events";
+import GitHubEvent from "@/components/gh_events/GitHubEvent";
+import { Octokit } from "octokit";
+import getGitHubAccessToken from "@/lib/getGitHubAccessToken";
+
+const GITHUB_PAT: string = getGitHubAccessToken() || "";
+const GITHUB_ORG: string = process.env.NEXT_PUBLIC_GITHUB_ORG || "";
```
rithviknishad, line 8: use typesafe `env` instead of `process.env`

app/feed/page.tsx
```diff
@@ -1,94 +1,54 @@
...
-import { useEffect, useState } from "react";
-import { scrollTo } from "@/lib/utils";
 import LoadingText from "@/components/LoadingText";
-import { env } from "@/env.mjs";
+import { IGitHubEvent } from "@/lib/gh_events";
+import GitHubEvent from "@/components/gh_events/GitHubEvent";
+import { Octokit } from "octokit";
+import getGitHubAccessToken from "@/lib/getGitHubAccessToken";
+
+const GITHUB_PAT: string = getGitHubAccessToken() || "";
```
rithviknishad, line 7: use pre-authenticated instance of octokit that is exported in recent commit: https://github.com/coronasafe/leaderboard/commit/9dd8348a0d29eadb2abee109af59d41699ae75d9#diff-d6bae8eb0c0c7b194e2579597b05c4f079d6e7dcdf48141956793f18b3361f21

app/feed/page.tsx
```diff
@@ -1,94 +1,54 @@
...
+export const revalidate = 600;
 
 type Props = {
   searchParams: {
     page?: string;
   };
 };
+const octokit = new Octokit({
+  auth: GITHUB_PAT,
+});
```
rithviknishad, line 19: use pre-authenticated octokit instance

https://github.com/coronasafe/leaderboard/commit/9dd8348a0d29eadb2abee109af59d41699ae75d9#diff-d6bae8eb0c0c7b194e2579597b05c4f079d6e7dcdf48141956793f18b3361f21

app/feed/page.tsx
```diff
@@ -1,94 +1,54 @@
...
+    {
+      org: GITHUB_ORG,
+      per_page: 1000,
+    },
+    (response) => {
+      const data = response.data as IGitHubEvent[];
+      const filteredData = data
+        .filter(exludeBotEvents)
+        .filter(excludeBlacklistedEvents);
+      return filteredData;
```
rithviknishad, line 33: Octokit responses are typesafe. Avoid casting it to another type.

```suggestion
      return response.data
        .filter(exludeBotEvents)
        .filter(excludeBlacklistedEvents);
```

app/feed/page.tsx
```diff
@@ -1,94 +1,54 @@
...
     </div>
   );
 }
 
-const exludeBotEvents = (event: IGitHubEvent) => {
-  return !event.actor.login.includes("bot");
-};
-
-const excludeBlacklistedEvents = (event: IGitHubEvent) => {
+const excludeBlacklistedEvents = (event: IGitHubEvent): boolean => {
```
rithviknishad, line 51: Why was this changed? Revert unrelated changes

image

rithviknishad (changes requested):

app/feed/page.tsx
dgparmar14: const data = response.data as IGitHubEvent[];
If we remove this line it will create a `Type error: No overload matches this call.`

rithviknishad: Previous code didn't use octokit either.

dgparmar14: Yes you are right , But in this if we remove this line so it's genrate an error on build time .
Also we are passing this in Githubevent  component as a prop to render and it also verify type using IGithubEvebt[]. So we have to also make changes in type.ts right?

rithviknishad: You won't need to define the types if you inline those functions right?

Existing types can be kept as is as it maybe used elsewhere.

rithviknishad (approved):
Awesome!

Issue #320, opened 2024-03-04
feature: Adding new container to show forked and starred count


Ashutosh0602:
This new feature will tell the new contributors the size of community and will encourage them to contribute and actively participate more in improving the CARE.

Will be adding this container as shown below in image within red reactangular box.

<img width="453" alt="Screenshot 2024-03-04 at 2 14 05 PM" src="https://github.com/coronasafe/leaderboard/assets/85889617/b66858c0-83f9-4ff7-90f2-d1d3df940865">

Ashutosh0602:
@rithviknishad I have created this issue as we discussed.
If it look okay to you assign me and I will raise PR.

PR #322, opened 2024-03-04
Added new Forked and Star count feature

Ashutosh0602:
close: #320 

This new feature will tell the new contributors the size of community and will encourage them to contribute and actively participate more in improving the CARE.

It will show the combine counts of forked and start count of all repositories of CARE at once.

Screenshot 2024-03-04 at 2 14 05 PM

### DIFF
--- a/app/feed/page.tsx
+++ b/app/feed/page.tsx
@@ -1,11 +1,12 @@
-"use client";
-
-import GitHubEvent from "@/components/gh_events/GitHubEvent";
-import { IGitHubEvent, combineSimilarPushEvents } from "@/lib/gh_events";
-import { useEffect, useState } from "react";
-import { scrollTo } from "@/lib/utils";
 import LoadingText from "@/components/LoadingText";
+import { IGitHubEvent } from "@/lib/gh_events";
+import GitHubEvent from "@/components/gh_events/GitHubEvent";
 import { env } from "@/env.mjs";
+import octokit from "@/lib/octokit";
+
+const GITHUB_ORG: string = env.NEXT_PUBLIC_GITHUB_ORG;
+
+export const revalidate = 600;
 
 type Props = {
   searchParams: {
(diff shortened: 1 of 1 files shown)


app/ForkedStar.tsx
```diff
@@ -0,0 +1,72 @@
...
+    `https://api.github.com/orgs/${env.NEXT_PUBLIC_GITHUB_ORG}/repos`,
+  );
+  const posts = await res.json();
+
+  return {
+    props: {
+      posts,
+    },
+  };
+}
```
rithviknishad, line 70: This does not work in app router.

Ashutosh0602:
@rithviknishad Is this okay or do you want to suggest some changes

app/ForkedStar.tsx
```diff
@@ -0,0 +1,72 @@
...
+  data?.map((ls: any) => {
+    if (ls.forks) {
+      forkCount += ls.forks;
+    }
+    if (ls.stargazers_count) {
+      starCount += ls.stargazers_count;
+    }
+    // console.log(ls.full_name, ls.forks, ls.stargazers_count);
+  });
+  console.log(forkCount);
```
rithviknishad, line 29: Let's make this as Server component instead of a client component

rithviknishad (changes requested):

app/ForkedStar.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <div className="flex justify-center">
+      {forkCount > 0 && starCount > 0 ? (
+        <>
+          <div className="flex items-center justify-center bg-violet-950 m-4 p-2 w-6/12 rounded-md border-2 border-slate-400 ">
```
rithviknishad, line 35: Let's not use colors outside `secondary` / `primary`. Different organizations deploying leaderboard would have different themes, and the CSS Variables secondary and primary uses the intended theme colors.

rithviknishad (changes requested):

app/ForkedStar.tsx
Ashutosh0602: Okay so it means I can use getServerSideProps to make it server sider rendering?

rithviknishad: Those are relevant to pages router only. These are app router pages.

Check out NextJS docs on App Router
https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#fetching-data-on-the-server-with-fetch

Ashutosh0602:
@rithviknishad I have made changes please review it.

app/ForkedStar.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <></>
+      )}
+    </div>
+  );
+}
+
+async function getData() {
+  const res = await fetch(
+    `https://api.github.com/orgs/${env.NEXT_PUBLIC_GITHUB_ORG}/repos`,
+  );
```
rithviknishad, line 50: 1. This fetches only first 30 repositories. 
2. Use the pre-authenticated `octokit` instance to make calls with GitHub APIs.
3. Use `octokit.paginate` to fetch all repositories from all pages.

app/ForkedStar.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { GoRepoForked } from "react-icons/go";
+import { GoStar } from "react-icons/go";
+
+export default async function ForkedStar() {
+  let forkCount = 0;
+  let starCount = 0;
+
+  const data = await getData();
+
+  data?.map((ls: any) => {
```
rithviknishad, line 11: Avoid usage of `any` types. This won't be needed as octokit responses are typesafe.

rithviknishad (changes requested):

PR #351, opened 2024-03-07
fixed -> clicking on commit link on /feed. redirects to GitHub API

manu-r12:
The actual problem was that the url to specifc commit was taking to the GitHub API not to the actual html url


components/DateRangePicker.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+`import React from "react";
```
rithviknishad, line 1: I assume this change was accidental?

```suggestion
import React from "react";
```

yarn.lock
rithviknishad, line 1: Remove this file, we are using pnpm lock file in this repo.

rithviknishad (changes requested):
Deploy failing...

image

yarn.lock
manu-r12: > Remove this file, we are using pnpm lock file in this repo.

which one?

manu-r12: Done ✅

manu-r12:
@rithviknishad  can you review it now 😃

rithviknishad (approved):

PR #353, opened 2024-03-07
Adds support for searching in /people page and show short info about user

### LINKED ISSUE #333: Adding search bar to `/people` and popover on hover to show contributor details
### Action Items

- [ ] Add search bar to `/people` page. Should search against their Name, GitHub username
- [ ] Upon mouse hovering on a contributor, a popover should appear showing details about the contributor.

Just a very rough sketch...

image

rash-27:
closes #333


rithviknishad (commented):
@rash-27 

1. It'd be great if you could give a meaningful title for the PR. Eg: "Adds support for searching in `/people` page and show short info about user"

2. Kindly link the issue in the PR's comment with a closing tag. Ref: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

lib/types.ts
```diff
@@ -161,6 +161,11 @@ export type LeaderboardPageProps = {
     ordering?: "asc" | "desc";
   };
 };
+export type ContributorsPageProps = {
+  searchParams: {
+    search?: string;
+  };
+};
```
rithviknishad, line 168: Let's keep the Page Props in the `page.tsx` itself instead as this wouldn't be reused. Only keep reusable types in this file

app/people/_components/Searchbar.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <div className="flex flex-col flex-wrap gap-4 md:flex-row">
+        <Search
+          defaultValue={searchParams.search}
+          handleOnChange={(e) => handleSearch(e.target.value)}
+          className="grow"
+        />
+      </div>
+    </div>
+  );
+}
```
rithviknishad, line 42: Can we make this a reusable component since leaderboard page also have this same component. Code duplication is hard to maintain.

app/people/_components/HoverInfoCard.tsx
rithviknishad, line 1: How about showing the contributors markdown bio also here?

app/people/_components/ContributorsPage.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <li
+              key={c.github}
+              className="transition-all duration-150 ease-in-out hover:scale-125 hover:shadow-xl hover:shadow-primary-500"
+            >
+              <Link href={`/contributors/${c.github}`}>
+                <Image
+                  onMouseEnter={() => {
+                    setIsHoverd(c.github);
+                  }}
+                  onMouseLeave={() => setIsHoverd("")}
```
rithviknishad, line 39: We could do this using tailwindcss classes itself right? Instead of using javascript?
Let's try to keep as minimal as client js as possible as this is "mostly" as static site.

app/api/leaderboard/functions.ts
```diff
@@ -129,3 +129,34 @@ export default async function fetchGitHubReleases(
...
+      github: contributor.github,
+      linkedin: contributor.linkedin,
+      slack: contributor.slack,
+      twitter: contributor.twitter,
+      joining_date: contributor.joining_date,
+      role: contributor.role,
+      courses_completed: contributor.courses_completed,
+    };
+  });
+};
```
rithviknishad, line 162: Let's not define this here + why are we doing this? We do not require all these information in the `/people` page right?

rithviknishad (changes requested):

app/people/_components/HoverInfoCard.tsx
rash-27: @rithviknishad but considering the card size , if the bio is long it will not be consistent i guess

rithviknishad: We can increase the size vertically though right?

Like how we have in GitHub

image

rash-27: okk , got it , ill change

rash-27: Should i also make the height to a max value and overflow the remaining bio as if the bio is long it is looking like this
Screenshot 2024-03-08 144138

rithviknishad: First of all, we should decrease the font-size of the rendered markdown content a lot 😅 
The normal text itself seems to be bigger than the contributors name.

Yup, we can specify a max-height for edge cases.

components/SearchBar.tsx
```diff
@@ -0,0 +1,150 @@
...
+    text,
+  }),
+);
+
+type SearchOptions = {
+  isSearch?: boolean;
+  isDaterange?: boolean;
+  isRoleFilter?: boolean;
+  isSort?: boolean;
+};
```
rithviknishad, line 31: I meant to make the "Search text box" alone to be reusable component, not the entire leaderboard filters.

rithviknishad:
Also error in deployment...
image

components/SearchBar.tsx
rash-27: Okay , I'll correct it

rithviknishad: yup!

PR #354, opened 2024-03-08
Migrate fetchGitHubReleases Method to Use Octokit for GitHub API Interactions

### LINKED ISSUE #347: Migrate `fetchGitHubReleases` method to use `octokit` instead of using fetch to interact with GitHub APIs
https://github.com/coronasafe/leaderboard/blob/main/app/api/leaderboard/functions.ts#L65

dgparmar14:
### Description
This pull request addresses the issue by migrating the fetchGitHubReleases method to use Octokit instead of the traditional fetch to interact with GitHub APIs. I have made several modifications as follows:

- Utilized `Octokit.graphql` for interacting with GitHub API.
- Replaced the conventional for loop with a `flatMap`.
- Adjusted the `ReleasesResponse` type interface to align with the response type from Octokit GraphQL.
- Install `@octokit/graphql`

Fixes #347 

### Type of change

1. Integration of `Octokit.graphql` for GitHub API interaction.
2. Implementation of `flatMap` for improved efficiency.
3. Refactoring of `ReleasesResponse` interface to match Octokit GraphQL response structure.

### DIFF
--- a/app/api/leaderboard/functions.ts
+++ b/app/api/leaderboard/functions.ts
@@ -1,6 +1,10 @@
-import { LeaderboardSortKey, ReleasesResponse } from "@/lib/types";
+import {
+  LeaderboardSortKey,
+  ReleasesResponse,
+  LeaderboardAPIResponse,
+  Release,
+} from "@/lib/types";
 import { getContributors } from "@/lib/api";
-import { Repository, LeaderboardAPIResponse, Release } from "@/lib/types";
 import { env } from "@/env.mjs";
 import { getGitHubAccessToken } from "@/lib/octokit";
 import octokit from "@/lib/octokit";
--- a/app/api/leaderboard/functions.ts
+++ b/app/api/leaderboard/functions.ts
@@ -81,7 +81,7 @@ export default async function fetchGitHubReleases(
                       login
                       avatarUrl
                     }
-                    mentions (first: 10) {
+                    mentions (first: 100) {
                       nodes {
                         login
                         avatarUrl
--- a/app/api/leaderboard/functions.ts
+++ b/app/api/leaderboard/functions.ts
@@ -1,13 +1,9 @@
-import { LeaderboardSortKey } from "@/lib/types";
+import { LeaderboardSortKey, ReleasesResponse } from "@/lib/types";
 import { getContributors } from "@/lib/api";
-import {
-  ReleasesResponse,
-  Repository,
-  LeaderboardAPIResponse,
-  Release,
-} from "@/lib/types";
+import { Repository, LeaderboardAPIResponse, Release } from "@/lib/types";
 import { env } from "@/env.mjs";
 import { getGitHubAccessToken } from "@/lib/octokit";
+import octokit from "@/lib/octokit";
 
 export const getLeaderboardData = async (
   dateRange: readonly [Date, Date],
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@headlessui/react": "^1.7.18",
+    "@octokit/graphql": "^8.0.1",
```
rithviknishad, line 16: Why was this package added? The default octokit package which we already have itself can dispatch graphql queries right? And this package doesn't seems to be used in the PR you've made either.

app/api/leaderboard/functions.ts
```diff
@@ -70,62 +70,57 @@ export default async function fetchGitHubReleases(
...
-    for (const release of repository.releases.nodes) {
-      release.repository = repository.name;
-      allReleases.push(release);
-    }
+      `,
+      org: env.NEXT_PUBLIC_GITHUB_ORG,
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        "Content-Type": "application/json",
+      },
```
rithviknishad, line 108: this is not required when using the pre-authenticated octokit instance since the credentials are already specified in the octokit construction

app/api/leaderboard/functions.ts
```diff
@@ -70,62 +70,57 @@ export default async function fetchGitHubReleases(
...
-                  description
-                  url
-                  author {
-                    login
-                    avatarUrl
-                  }
-                  mentions (first: 10) {
-                    nodes {
-                      login 
+  try {
```
rithviknishad, line 73: let's not wrap it in a try-catch. Let the build fail if there is an error, since this would be taken care during pre-rendering.

rithviknishad (changes requested):

app/api/leaderboard/functions.ts
```diff
@@ -65,67 +64,48 @@ export const getLeaderboardData = async (
...
-                    avatarUrl
-                  }
-                  mentions (first: 10) {
-                    nodes {
-                      login 
+  const response: ReleasesResponse = await octokit.graphql({
+    query: `
+        query GetReleases($org: String!) {
+          organization(login: $org) {
+            repositories(first: 100) {
```
rithviknishad, line 71: order by updated at descending

rithviknishad (changes requested):

app/api/leaderboard/functions.ts
dgparmar14: image

It's not supporting UPDATED_AT ,
I think we have to sort the fetched releases based on their updated at field.?

rithviknishad: No, it is supported. I meant to sort repositories by updated at, not releases. releases can remain created at itself.

dgparmar14: image

Is it fine okk?

rithviknishad: yup

rithviknishad (approved):

rithviknishad (changes requested):
Can we increase the number of contributors shown per release limit to something like a 100?

This release had lots of contributors but we are showing incorrect number of contributors here...
image


We can reduce the number of repositories down to first 50 maybe if github warns about node limit exceeded.

dgparmar14:
If we reduce the number of repositories down to the first 50 and we increase the number of contributors shown per release limit to something like 100. It looks like this
image
it's looks something like this.
And if we  try to fetch more than 100 mentions whether repo is 50 or 20 github will not allowed

rithviknishad (approved):

Issue #346, opened 2024-03-06
Migrate `GitHubEvents` component to use `octokit` instead of using fetch to interact with GitHub APIs


rithviknishad:
https://github.com/coronasafe/leaderboard/blob/main/components/gh_events/GitHubEvents.tsx

Ashutosh0602:
@rithviknishad I would like to work on this issue

PR #365, opened 2024-03-10
Replaced fetch with Ocktokit in GitHubEvents

Ashutosh0602:
Removed fetch and used Ocktokit to retrieve data from API in `GitHubEvents.tsx` file.

 fix: #346


Ashutosh0602:
@rithviknishad I have tried to replace fetch ocktokit is it right?

components/gh_events/GitHubEvents.tsx
```diff
@@ -57,3 +51,13 @@ export default function GitHubEvents({ minimal }: { minimal?: boolean }) {
     </div>
   );
 }
+
+async function getEvent() {
+  const response = await octokit.paginate("GET /orgs/{org}/events", {
```
rithviknishad, line 56: We only need the first 5 events. This does not require us to paginate as the first page of 100 results would be sufficient.

```suggestion
  const response = await octokit.request("GET /orgs/{org}/events", {
```

components/gh_events/GitHubEvents.tsx
```diff
@@ -57,3 +51,13 @@ export default function GitHubEvents({ minimal }: { minimal?: boolean }) {
     </div>
   );
 }
+
+async function getEvent() {
```
rithviknishad, line 55:
```suggestion
async function getEvents() {
```

components/gh_events/GitHubEvents.tsx
```diff
@@ -21,29 +20,24 @@ const excludeBlacklistedEvents = (event: IGitHubEvent) => {
...
-            data.filter(exludeBotEvents).filter(excludeBlacklistedEvents),
-          ).slice(0, 5),
-        ),
-      );
-  }, [page]);
+export default async function GitHubEvents({ minimal }: { minimal?: boolean }) {
+  const events = await getEvent().then((data) =>
+    data
+      .filter(excludeBlacklistedEvents as any)
+      .filter(exludeBotEvents as any)
```
rithviknishad, line 27: do not use `any` instead try inlining the function so that you can skip defining the types.

components/gh_events/GitHubEvents.tsx
```diff
@@ -21,29 +20,24 @@ const excludeBlacklistedEvents = (event: IGitHubEvent) => {
...
+      .slice(0, 5),
+  );
+  if (!env.NEXT_PUBLIC_GITHUB_ORG) {
+    throw new Error("Missing NEXT_PUBLIC_GITHUB_ORG");
+  }
```
rithviknishad, line 32: Let's not do this here. Ideally, this should be done at by env schema validation at build time and not during rendering. 

```suggestion
```

components/gh_events/GitHubEvents.tsx
```diff
@@ -21,29 +20,24 @@ const excludeBlacklistedEvents = (event: IGitHubEvent) => {
...
 
   return (
     <div className="flow-root">
       <ul role="list" className="-mb-8 mt-4 flex flex-col gap-4 space-y-4">
         {events ? (
-          events.map((e) => <GitHubEvent key={e.id} event={e} />)
+          events.map((e) => (
+            <GitHubEvent key={e.id} event={e as IGitHubEvent | undefined} />
```
rithviknishad, line 39: just directly pass `e` without casting. If type definition at GitHub event is wrong, change that instead.

rithviknishad (changes requested):

Ashutosh0602:
@rithviknishad I made the changes have a look

components/gh_events/GitHubEvents.tsx
```diff
@@ -21,23 +20,13 @@ const excludeBlacklistedEvents = (event: IGitHubEvent) => {
...
-      .then((data) =>
-        setEvents(
-          combineSimilarPushEvents(
-            data.filter(exludeBotEvents).filter(excludeBlacklistedEvents),
-          ).slice(0, 5),
-        ),
-      );
-  }, [page]);
+export default async function GitHubEvents({ minimal }: { minimal?: boolean }) {
+  const events = await getEvents().then((res: any) =>
```
rithviknishad, line 24: Let's try to not use `any` type.

rithviknishad (changes requested):

rithviknishad (changes requested):
You need to ensure there are no errors before requesting for review. By removing `any` type, I meant to fix the type errors along with it.

image

PR #368, opened 2024-03-10
Enhance Release Event Card for Mobile View closes #358

### LINKED ISSUE #358: Enhance Release Event Card for Mobile View
https://contributors.ohc.network/releases

image

rash-27:
- closes #358 
- if width is above 390px the care/org is shown
Screenshot from 2024-03-11 01-10-01
- if width is below 390px it is hidden
Screenshot from 2024-03-11 01-08-52

### DIFF
--- a/app/releases/page.tsx
+++ b/app/releases/page.tsx
@@ -3,9 +3,9 @@ import Image from "next/image";
 import Markdown from "@/components/Markdown";
 import Link from "next/link";
 import { FiGithub } from "react-icons/fi";
-import { env } from "@/env.mjs";
+import { FaRegStar } from "react-icons/fa";
 import { formatDate } from "@/lib/utils";
-
+import { GoTag } from "react-icons/go";
 export const revalidate = 900; // revalidates atmost once every 15 mins
 
 export default async function Page() {
@@ -18,6 +18,11 @@ export default async function Page() {
       </span>
     );
   }
+  // Finds all the repositories present
+  let myRepos = new Set<string>();
+  releases.map((release) => {
+    myRepos.add(release.repository);
+  });
 
   return (
     <div>
@@ -29,10 +34,21 @@ export default async function Page() {
           >
             <div className="flex items-center justify-between p-3 pb-0 pt-4 sm:p-6">
               <div className="flex items-center">
-                <h3 className={`font-semibold`}>
-                  {release.repository}
-                  {" - "}
-                  {release.name}
+                <h3 className={`flex justify-center font-semibold`}>
+                  <div className="hidden sm:inline">{release.repository}-</div>
+                  <GoTag
+                    className="relative pt-1.5"
+                    size={35}
+                    color="#238636"
+                  />
+                  <div className="px-1 ">{release.name}</div>
+                  {myRepos.has(release.repository) &&
+                    myRepos.delete(release.repository) && (
+                      <FaRegStar
+                        className="relative mx-2 pt-1 "
+                        color="#238636"
+                      />
+                    )}
                 </h3>
               </div>
               <a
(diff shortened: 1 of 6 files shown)


app/releases/page.tsx
```diff
@@ -50,19 +50,25 @@ export default async function Page() {
...
+                  className="text-sm font-semibold sm:text-base"
                 >
                   {release.author.login}
                 </a>{" "}
+                on{" "}
+                {new Date(release.createdAt).toLocaleDateString("en-US", {
+                  day: "2-digit",
+                  month: "short",
+                  year: "numeric",
+                })}
```
rithviknishad, line 67: Use the custom `formatDate` utility function instead, so that we have consistent date formats across the site.

rithviknishad (changes requested):
Can we move the repo name to the top to the left of open in github button? We can keep icon only for open in github in mobile view.

Also, it is totally fine to remove the github org from the card in desktop view too if it is making things complicated to achieve it

image

app/releases/page.tsx
rash-27: okk I will make changes and update you , and I have used that format as it is also used in the Release section in homepage

rithviknishad: nice find, update there too 😄

rash-27: okk , ill update it

rash-27: image
image

how does this look ? i have displayed only the git hub icon when dimension is less than 370px , and also I have one doubt like is it not necessary to put the version number as we do previously before shifting to octokit ?

rithviknishad: Yes, it is important to show the version number.

rash-27: Currently i guess it is not fetched , can i take up this issue and solve ?

rithviknishad: yup

rash-27: Should i solve that part too under this issue ?

rithviknishad: Create a new issue, but could be solved in the same PR by linking multiple issues to it

rash-27: okk

rash-27:
is it ok ?
image

rithviknishad:
Yup Nice! can we also add a tag icon before the version?

rithviknishad:
Also, can we mark the latest release of a particular repo with the "latest" tag/chip?

rash-27:
Yeah sure , I'll work on it

rash-27:
@rithviknishad I couldnt find the latest tag in react-icons , so i have added a star as shown for latest release in a particular repository , how does this look ?
image

rithviknishad:
This looks nice, but are we not showing the repository name? 👀

rithviknishad:
I think, just specifying "Latest" would be better than trying to use an icon. A user may not recognize that this is the latest by looking at the star.

rash-27:
In phone view I am only displaying the version , but in remaining both are displayed

rithviknishad:
It is okay to hide org name in both views. But I don't think it is okay to hide the repository name. An organization may have multiple products/repositories each having its releases. One would not be able to distinguish which version belongs to which repository by looking at this from mobile right?

This would've been fine only if the releases page was specific to a repository.

rash-27:
okk , then ill display just the repository name in phone view , is it sufficient ?
image
and here is the latest tag here ok ?

rithviknishad:
Lets show both repo name and version name in mobile view. It's fine to make the text smaller or put things in newline in mobile view

rash-27:
Ok , I'll update it

app/releases/page.tsx
```diff
@@ -26,43 +31,52 @@ export default async function Page() {
...
+                    className="relative hidden pt-1 text-lg sm:inline sm:pt-1.5 "
+                    size={30}
+                    color="green"
+                  />
+                  {release.name}
+                  {myyRepos.has(release.repository) &&
+                    myyRepos.delete(release.repository) && (
+                      <span className="relative ml-2 rounded-full border border-green-700 p-1 pt-2.5 text-xs text-green-700 ">
+                        Latest
+                      </span>
```
rithviknishad, line 51: the "latest" chip has improper padding. this does not look nice...

image


```suggestion
                      <span className="relative ml-2 rounded-full border border-green-700 p-1 text-xs text-green-700 ">
                        Latest
                      </span>
```

rithviknishad (approved):

PR #374, opened 2024-03-11
 Adds support for searching in /people page and show short info about user

### LINKED ISSUE #333: Adding search bar to `/people` and popover on hover to show contributor details

rash-27:
- closes #333


rash-27:
@rithviknishad for markdown component can we send some kind of props (for size ) as it is used at many places and needed to be rendered at different sizes  ?

components/Markdown.tsx
```diff
@@ -14,7 +17,10 @@ export default async function Markdown(props: { children: string }) {
 
   return (
     <div
-      className="prose font-inter text-xs leading-relaxed dark:prose-invert md:text-sm xl:text-left"
+      className={clsx(
+        `prose font-inter leading-relaxed dark:prose-invert xl:text-left`,
+        props.className ? props.className : "text-sm sm:text-base",
+      )}
       dangerouslySetInnerHTML={{ __html: result.toString() }}
     />
```
rithviknishad, line 25: I'd suggest to do something like the one shown, so that when someone tries to override the bg-color let's say, it'd cause a weird behaviour the default text size classes are no longer applied and would cause confusion. Doing the below method would eliminate that.

```jsx
<div className="default classes...">
  <div 
    classsName={props.className}
    dangerouslySetInnerHTML={{ __html: result.toString() }}
  />
</div>
```

rithviknishad (changes requested):

components/Markdown.tsx
rash-27: Yeah that sounds great

bodhish:
It would be nice to show a search animation when we start the network request, currently its hard to notice when the search starts;

rash-27:
@bodhish, I didn't understand what you mean by showing an animation when the network search starts. Do we need to show a search animation during the delay due to debouncing?

bodhish:
The UI should show search happening

rash-27:
Yeah got it ,I will update it , thank you

rash-27:
- Before
Screenshot from 2024-03-31 23-46-15
- After
Screenshot from 2024-04-01 00-03-31
[Screencast from 01-04-24 12:01:41 AM IST.webm](https://github.com/coronasafe/leaderboard/assets/146672184/367518b7-96c2-4879-bedd-247436255864)

rithviknishad (approved):
LGTM. We could have made the footer stay at the bottom persistently. But can be a separate PR.

rash-27:
Ok , we can link this issue with the color of the footer during the theme change.

PR #376, opened 2024-03-11
fix github og images where repo starts with a period

sainak:
Fixes the issue where the period in the path will fail to match the github opengraph remotePatterns

this resulted in a crash on local:
image

and failed to generate image on deployment:
image

after fix:
image


rithviknishad (commented):
LGTM Can we also add comments here on why we are doing this?

rithviknishad (approved):

Issue #377, opened 2024-03-12
No Hyperlink support for comments 


UdaySagar-Git:
- Currently, there is hyperlink support for all other types except for 'comment_created.' It would be better to have hyperlinks instead of plain text so that one can easily click on the text to get redirected to the comment section.

![image](https://github.com/coronasafe/leaderboard/assets/111575806/214e8925-c01f-47fe-9bad-0c39c80cb92f)

UdaySagar-Git:
can you please assign this to me

PR #380, opened 2024-03-12
Added hyperlink for comments, and changed from using Link to anchor tag

UdaySagar-Git:
Fixes - #377 

- Changed from using Link to anchor tags , since those are not used for client side navigation and are external links, and would be better if they open in new tab by default  

image

### DIFF
--- a/components/contributors/GithubActivity.tsx
+++ b/components/contributors/GithubActivity.tsx
@@ -134,12 +134,8 @@ let renderText = (activity: Activity) => {
                 <RelativeTime time={timestamp} />
               </span>
             </div>
-            <div className="mt-2 rounded-lg border border-secondary-600 p-2 md:p-4">
-              <a href={activity["link"]} target="_blank">
-                <span className="cursor-pointer break-words text-sm font-medium text-foreground hover:text-primary-500">
-                  {activity["text"]}
-                </span>
-              </a>
+            <div className="max-w-xl pt-4">
+              <OpenGraphImage url={activity["link"]} className="rounded-xl" />
             </div>
           </div>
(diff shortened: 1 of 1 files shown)


rithviknishad (changes requested):
Nice work... Can we also comment look like a comment by adding a border around it and improving the UI?

Currently this does not look/feel like a comment

UdaySagar-Git:
@rithviknishad sure

UdaySagar-Git:
@rithviknishad does this look fine?
image

rithviknishad:
Yup, maybe we can add a bit more padding in desktop view and reduce the font size a bit.

UdaySagar-Git:
@rithviknishad 
I think we need to add that to all type of events ?
what do you think ? is it ok to change ui only for comments
image

rithviknishad:
Yea, but for pull requests open/merge, or issue open/close, review approvals without comments, we should be using the OpenGraphImage. Rest of the comment events can be done like as you've mentioned

UdaySagar-Git:
@rithviknishad 
image

Should i add the styles for issue opened /closed ??

rithviknishad:
@UdaySagar-Git why are we changing those?

We should ideally be showing `<OpenGraphImage />` for those instead like how we have in feed.
Was it not the case already?

image

UdaySagar-Git:
@rithviknishad we are not showing OpenGraphImage  earlier , 
i will change that to use openGraphImage

UdaySagar-Git:
@rithviknishad can you please review now !

rithviknishad (approved):

PR #393, opened 2024-03-17
Implement API endpoints for tasks bot slack app

rithviknishad:
### Proposed Changes

- Implements API endpoints required for basic CRUD operations for managing tasks from Slack Bot
- Uses Vercel's KV Redis Storage for storing the tasks for each slack user

https://github.com/coronasafe/leaderboard/assets/25143503/94f8a3e3-a5a9-4a46-89af-c56e48597128

### DIFF
--- a/app/api/task/complete/route.ts
+++ b/app/api/task/complete/route.ts
@@ -1,0 +1,20 @@
+import { parseSlackRequest } from "@/app/api/task/slack";
+import Tasks from "@/app/api/task/task";
+
+export async function POST(request: Request) {
+  const { userId, text } = await parseSlackRequest(request);
+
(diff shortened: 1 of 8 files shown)


Ashesh3 (approved):
LGTM, by the way I seem to be missing the context for having Todo manager endpoints in the leaderboard repo 😅 What was the use case for this?

Also, is the TODO for a later stage?
`TODO: Ensure incoming requests are from allowed sources`

rithviknishad:
@Ashesh3 use case was simply to manage tasks at an org. level. We had discussed it to keep it in leaderboard for now, and if the requirement becomes larger, we can move it to a separate repo.

About the TODO, yes. Will add security header check...

PR #409, opened 2024-03-29
Scaling points for PRs based on size of the PR

### LINKED ISSUE #219: Scaling points for PRs based on size of the PR / complexity of the issue
## Problem

Although the leaderboard tracks the pull requests that are made, it currently does not track how big or small the PR is.

## Proposed Solution

Track the number of linked issues to the PR and increase the point by 1 for each linked issue.

Samuel-Aktar-Laskar:
Closes #219 

Update in Scraper (github.py)
I checked the count of linked issues to a pr from pr body and added field "no_of_linked_issue" to the event. 

In the api.ts file, I adjusted the code account for no of linked issues. So now total point for a PR would be (1 for opening pr + no of linked issues).


scraper/src/github.py
```diff
@@ -125,6 +135,7 @@ def parse_event(self, event, event_time):
                         "time": event_time,
                         "link": event["payload"]["pull_request"]["html_url"],
                         "text": event["payload"]["pull_request"]["title"],
+                        "no_of_linked_issues" : no_of_linked_issues
```
rithviknishad, line 138: can we store the references to the issues (html_url) itself directly here?

scraper/src/github.py
```diff
@@ -116,6 +124,8 @@ def parse_event(self, event, event_time):
                 )
 
         elif event["type"] == "PullRequestEvent":
+            pr_body = event["payload"]["pull_request"]["body"]
+            no_of_linked_issues = self.parse_linked_issues(pr_body)
             if event["payload"]["action"] == "opened":
```
rithviknishad, line 129: Problem with this approach is that, we parse the number of linked issues for PR open events.

But during the lifecycle of a PR, we may add/remove the issues linked with the PR before it gets merged. Those wouldn't be reflected here...


We can scale points for merged pr's alone, not for pr's opened.

scraper/src/github.py
```diff
@@ -71,6 +71,14 @@ def append(self, user, event):
                 "authored_issue_and_pr": [],
             }
 
+    def parse_linked_issues(self, pr_body):
+        if isinstance(pr_body, str):
+            pattern = r"#(\d+)"
+            matches = re.findall(pattern, pr_body)
```
rithviknishad, line 77: 1. Issues can be linked to a PR even without having to mention the issue number in the PR body. Those may not be captured this way.

2. We may reference an issue in the PR body, but it may not necessarily solve the issue. Use actual.


Refer all the ways in which link an issue to the PR here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

lib/api.ts
```diff
@@ -248,29 +268,30 @@ function getCalendarData(activity: Activity[]) {
...
+//   return HIGHLIGHT_KEYS.map(
+//     (key) => points[key] * (calendarDataEntry[key] ?? 0),
+//   ).reduce((a, b) => a + b, initialPoints);
+// };
 
 const HighlightsReducer = (acc: Highlights, day: Highlights) => {
   return {
-    points: computePoints(day, acc.points),
+    // points: computePoints(day, acc.points),
+    points: acc.points + (day.points ?? 0),
```
rithviknishad, line 294: instead of changing this, tweak the `computePoints` instead.

Let's make computePoints more generic (being able to compute points dynamically based on conditions etc. instead of simply having a lookup to the `points` map).

lib/api.ts
```diff
@@ -218,6 +228,16 @@ function getCalendarData(activity: Activity[]) {
       } else {
         acc[date][activity.type] = 1;
       }
+      if (!acc[date]["points"]) {
+        acc[date]["points"] = points[activity.type];
+      } else {
+        acc[date]["points"] += points[activity.type];
+        if (activity.type === "pr_opened") {
+          acc[date]["points"] +=
+            (activity.no_of_linked_issues ?? 0) * points["no_of_linked_issues"];
```
rithviknishad, line 237: like mentioned, do for pr's merged instead

lib/api.ts
```diff
@@ -15,6 +15,7 @@ const points = {
   issue_opened: 4,
   eod_update: 2,
   pr_opened: 1,
+  no_of_linked_issues: 1,
```
rithviknishad, line 18: wouldn't be necessary when compute_points method is tweaked...

```suggestion
```

rithviknishad (changes requested):

Samuel-Aktar-Laskar:
@rithviknishad 
I have created separate file for parser where I collect the linked issues from pr body description, commits messages and UI linked issue. For UI linked issue, I am using graphql (https://stackoverflow.com/a/61940472). Let me know if there is any better way to do. I tested the same by making demo issue and prs in my own account, and it seems to work fine. 

Rest are the changes in sorting logic. 
Let me know the changes or improvements required.

rithviknishad:
@Samuel-Aktar-Laskar 

We have moved away from python based github scraper and we are now using octokit in typescript to achieve the same. We can use octokit's graphql fn. to perform graphql calls to scrape data too! 

Feel free to re-open a fresh PR 😄

PR #410, opened 2024-03-30
Fixes issue with current level resolver for GA if not achieved

rithviknishad:
Fixes issue with current level resolver for graduate attributes taking first level automatically if none achieved.

**Before:**
image

**After:**
image

cc: @bodhish


bodhish (approved):
LGTM

PR #414, opened 2024-03-30
Updated README.md

Harshal141:
- Added Prerequisites section.
- command to run the server using contributors GitHub access key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,7 @@ A simple leaderboard app built with Next.js and Tailwind CSS to list the top con
 
 ### Prerequisites
 
-- Ensure that `npm` is installed on your device, or you can use a package manager like `nvm` to manage multiple versions of Node.js. [Read more](https://www.educative.io/answers/what-is-nvm)
-- Install pnpm, a package manager, to run the project.
-```bash
-npm install -g pnpm
-```
+Ensure that `pnpm` is installed on your device. You can check the steps for installation here. [Docs Link](https://pnpm.io/installation)
 
 ### Starting the development server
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,17 @@
...
 ## Getting Started
 
+### Prerequisites
+
+- Ensure that `npm` is installed on your device, or you can use a package manager like `nvm` to manage multiple versions of Node.js. [Read more](https://www.educative.io/answers/what-is-nvm)
+- Install pnpm, a package manager, to run the project.
+```bash
+npm install -g pnpm
+```
+
```
rithviknishad, line 14: `pnpm` can be installed in many ways and one need not necessarily need to have `npm`/`nvm` to install `pnpm`. Hence these need not be mentioned here, instead you can simply reference to the pnpm installation docs: https://pnpm.io/installation

README.md
```diff
@@ -20,6 +28,27 @@ pnpm dev
...
+- For new contributors you may encounter the error
+  ```
+   GITHUB_PAT is not configured in the environment.
+   Request quota exhausted for request POST /graphql
+  ```
+  In place of running `pnpm dev` create you own GitHub access token. [Read Steps here](https://docs.github.com/en/enterprise-server@3.9/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
+  Then run the following command instead:
+  ```
+  GITHUB_PAT=<YOUR_KEY> pnpm dev
+  ```
```
rithviknishad, line 43: You can also mention about using auth token provided by github's cli if user has gh cli installed that way contributor need not create a PAT token...

image

README.md
```diff
@@ -20,6 +28,27 @@ pnpm dev
...
+
+- If you are getting the error:
+  ```
+   fatal: unable to access 'https://github.com/coronasafe/leaderboard-data.git/': Could not resolve host: github.com
+   Error executing command: git clone
+  ```
+  Make sure you are connected to an active internet connection, or your antivirus may also cause this issue. Try running the command with your antivirus closed.
```
rithviknishad, line 50: Should we mention this? I doubt if they can clone this repo itself if contributor has these issues. It could also be because of some networks blocking github too, there are far too many reasons than the ones mentioned for this to happen.

rithviknishad (changes requested):

README.md
Harshal141: When I cloned the main repo there was no issue with the installation and running. 
But when I tried running pnpn install. It was showing an error for cloning the data repo. Only after I closed my antivirus the download was successful.
I think this should be mentioned or if you want we can remove it.

README.md
Harshal141: Ok got it. This is better than directly putting the key.

README.md
Harshal141: I thought most people have npm that's why we can use this method.
But your point is also right I will update this.

README.md
Harshal141: I have removed it

rithviknishad (approved):
LGTM

PR #418, opened 2024-04-01
enable/disable features from env variables

### LINKED ISSUE #413: We should have an option to make the pages optional;
Few projects may not have a releases dashboard or projects board. It would have been nice if we had an option to enable/disable these from env

rash-27:
- fix #413


rithviknishad (changes requested):
Can we create a typesafe utility function that can be invoked like this to check the status of the feature flag?

```ts
if (!featureIsEnabled("feature-name")) {
  return notFound();
}
```

rithviknishad:
> I have considered the env variable as a string with features separated with a ', ', Is that a good practice to do so?

It should be fine for now. We can refactor it later if we encounter an issue with this approach.

> And should I also add checks for the leaderboard and other routes as well?
The leaderboard is always going to be present since all orgs would have some activity where the leaderboard would have data. Let's not disable it...
We can now have flags for routes like "projects" and "releases" for now.

cc: @bodhish

rash-27:
Yes , I'll update it

lib/utils.ts
```diff
@@ -93,3 +93,7 @@ export const navLinks = [
 export const formatDate = (date: Date) => {
   return format(date, "MMM dd, yyyy");
 };
+
+export const featureIsEnabled = (feature: string) => {
+  return process.env.NEXT_PUBLIC_FEATURES?.split(",").includes(feature);
+};
```
rithviknishad, line 99: import the typesafe env instead + add typesafety.

```suggestion
type Feature = "releases" | "projects";

export const featureIsEnabled = (feature: Feature) => {
  return env.NEXT_PUBLIC_FEATURES?.split(",").includes(feature);
};
```

rithviknishad (changes requested):

rithviknishad (changes requested):
LGTM, update the README too.

rash-27:
Ok sure!

rithviknishad (approved):

PR #452, opened 2024-05-09
Initialise NPM project for activities bot and lock dependency versions

rithviknishad:
### Changes

- Initialized an NPM Project for Activities Bot and locked dependency versions

### Reason

- Due to a recent release from octokit: https://github.com/octokit/octokit.js/releases/tag/v4.0.0 which broke our existing action [(run)](https://github.com/coronasafe/leaderboard-data/actions/runs/9002742523)


khavinshankar (approved):
LGTM,

one suggestion, why not separate the reminder and post update into separate workflows, the current, combined workflow runs for more than 30mins right?

rithviknishad:
@khavinshankar We had discussed this already, but making a single workflow may cause issues in some days as GitHub's CRON schedule trigger has lower tolerance.

So to avoid situations where the reminder may be accidentally posted after the Send to Slack, we decided to simply sleep for 30 minutes and run the next step.

Ref: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule


image

PR #456, opened 2024-05-18
EOD Reminders: Adds support for retries

### LINKED ISSUE #455: Add support for retries to slack post-message calls.
image

rithviknishad:
- Fixes #455

### DIFF
--- a/activities-bot/src/utils.mjs
+++ b/activities-bot/src/utils.mjs
@@ -44,7 +44,9 @@ function isAllowedEvent(event) {
   }
 }
 
-const throwForHttpError = async (res) => {
+const throwForHttpError = async (promise) => {
+  const res = await promise;
+
   if (!res.ok) {
     throw new Error(await res.text());
   }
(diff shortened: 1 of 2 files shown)


activities-bot/src/utils.mjs
```diff
@@ -252,3 +252,17 @@ export async function postEODMessage({ github, slack, updates }) {
     getHumanReadableUpdates(updates, slack, github),
   );
 }
+
+export async function withRetry(method, { attempts }) {
+  while (attempts) {
+    try {
+      return await method();
+    } catch (error) {
+      attempts -= 1;
```
khavinshankar, line 261: For the retries to work, do add a throw statement inside fetch when !ok, or lets create a wrapper around fetch which does that

rithviknishad: Done. Original issue was for non HTTP errors. This should now cover for HTTP errors too.

khavinshankar (approved):

PR #463, opened 2024-07-12
GSoC 2024: GitHub Discussion UI at Home,  Discussion and Contributor's route

### LINKED ISSUE #212: Support for showing GitHub discussions
## Project Detail

Leaderboard collects data from GitHub and Slack to show off the work of our open source contributors

**Is your feature request related to a problem? Please describe.**

Currently, we are not tracking GitHub discussions of an organization. It'd be great if we could show discussion events happening in the org. in leaderboard.

## Features To Be Implemented

- [ ] Add support to the scraper script to be able to scrap using github's graphql endpoint because discussion events can be accessed only via graphql.
- [ ] Scrapper should scrap discussion events and all participants of each discussions would have a reverse relation to the discussion event.
- [ ] Show latest 5 discussions in the leaderboard home page.
- [ ] A show more button should land to a larger page where all discussion events of the organization are shown.
- [ ] Contributor profile page should contain the discussion event if they are a participant or the author of the discussion.
- [ ] #68 

## Learning Path

|  | Details |
| :---:  | :---:  |
| Complexity       | Medium |
| Required Skills | NextJS, TypeScript, Python |
| Mentors            | @bodhish , @khavinshankar , @rithviknishad , @gigincg , @Ashesh3 , @sainak |
| Project Size      | 90 hours |

## Link to documentation for Product Set-Up

1. Refer to the readme file available at: https://github.com/coronasafe/leaderboard/blob/main/README.md

## Acceptance Criteria

- All the CI/CD workflows should pass (including the scraper workflow)
- Code is logically structured for long-term project sustainability.
- Proper descriptions should be included for review purposes.
- QA and Code Review

## Milestone

- [ ] Get to know Leaderboard as a tool and understand the workflow.
- [ ] Understand the requirements and come up with a design.
- [ ] Design and implement the updated UI
- [ ] Complete functionality of scraping and showing GitHub discussion events
- [ ] Optimise and improve UX
- [ ] Review and QA

dgparmar14:
# Pull Request Description

## Overview

This PR introduces a comprehensive update to display GitHub Discussions across various routes: Home, Discussions, and Profile. It also adds functionality to filter discussions based on their category and creation date. Additionally, discussions will be integrated with activities on the contributor's route.

Fixes #212 

### New Routes and Components
1. **New Discussions Route**:
   - Created a new route to display all discussions (currently limited to GitHub Discussions).

2. **New Components**:   
   - **GitHub Discussions**: Displays all discussions and handles their filtering.
   - **GitHub Discussion**: A discussion card showcasing the title, author, timeline, description, and participants.
   - **Filter Discussions**: Filters discussions using SearchParams. Includes:
     - **Discussions Category**
     - **Date Range Picker** (reuses the existing component)
   - **DiscussionsMarkdown**: A client-side markdown component for experimental purposes. It is intended to potentially replace the existing server-side component if proven effective.

### Library Update
- **lib/discussion.ts**: Contains two functions:
  - `fetchGithubDiscussionForUser()`
  - `fetchGithubDiscussion()`: Fetches a specified number of discussions. If no number is provided, it returns all discussions.


components/discussions/GithubDiscussions.tsx
```diff
@@ -0,0 +1,67 @@
...
+        );
+      });
+    } else if (filter) {
+      filterData = discussions.filter(
+        (discussion) => discussion.category?.name === filter,
+      );
+    }
+
+    setFilterDiscussions(filterData);
+  }, [filter, discussions, dateRange]);
```
rithviknishad, line 46: Let's not use a state here? We can achieve this without using a state too right?

We could directly filter the `discussions` outside the useEffect.

lib/api.ts
```diff
@@ -4,12 +4,15 @@ import { Activity, ActivityData, Contributor, Highlights } from "./types";
...
+import { fetchGithubDiscussionForUser } from "@/lib/discussion";
+import { ParsedDiscussion } from "@/scraper/src/github-scraper/types";
+import { env } from "@/env.mjs";
 
 const root = join(process.cwd(), "data-repo/contributors");
 const slackRoot = join(process.cwd(), "data-repo/data/slack");
 const githubRoot = join(process.cwd(), "data-repo/data/github");
 
-const points = {
+const points: { [key: string]: number } = {
```
rithviknishad, line 15: A cleaner more safer way is to use `as const` in the end. It'll have much more stricter type safety! 

Ref: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions


```js
const foo = {
  a: 'hello',
  b: 'world',
} as const;
```

```suggestion
const points = {
```

lib/api.ts
```diff
@@ -85,6 +88,24 @@ export async function getContributorsSlugs(): Promise<{ file: string }[]> {
...
+        ? `Started a Discussion`
+        : `Commented on a Discussion`;
+    return {
+      type: "discussion",
+      title: title,
+      time: discussion.time,
+      link: discussion.link,
+      discussion: discussion,
+    };
+  });
```
rithviknishad, line 105: Let's keep sentence constructions in the place where the page is rendered.

```suggestion
    return {
      type: "discussion",
      title: discussion.author === githubHandle ? "authored" : "commented",
      time: discussion.time,
      link: discussion.link,
      discussion: discussion,
    } as const;
  });
```

lib/discussion.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    user &&
+    discussions[0].filter(
+      (discussion: ParsedDiscussion) =>
+        (discussion.participants ?? []).includes(user) ||
+        discussion.author === user,
+    )
+  );
+}
```
rithviknishad, line 56: Avoid code duplication. This fn. does everything same as `fetchGithubDiscussion` in addition to just filtering by user right? Compose the fn. to reuse the existing fn. and just add a filter instead of duplicating the file read logic.

lib/types.ts
```diff
@@ -66,16 +67,18 @@ export const ACTIVITY_TYPES = [
...
+  "discussion",
 ] as const;
 
 export interface Activity {
   type: (typeof ACTIVITY_TYPES)[number];
   title: string;
   time: string;
   link: string;
-  text: string;
+  text?: string;
```
rithviknishad, line 78: Could you specify why this was made nullable?

lib/utils.ts
```diff
@@ -88,12 +88,13 @@ export const navLinks = [
   { title: "People", path: "/people" },
   { title: "Projects", path: "/projects" },
   { title: "Releases", path: "/releases" },
+  { title: "Discussion", path: "/discussion" },
```
rithviknishad, line 91: Update the [`sitemap.ts`](https://github.com/coronasafe/leaderboard/blob/main/app/sitemap.ts) too to include this new page

rithviknishad (changes requested):

components/DscMarkdown.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }, [props.children]);
+  return (
+    <div className="prose font-inter text-sm leading-relaxed dark:prose-invert sm:text-base xl:text-left">
+      <div
+        className={clsx(props.className ?? "")}
+        dangerouslySetInnerHTML={{ __html: processedContent.toString() }}
+      />
+    </div>
+  );
+}
```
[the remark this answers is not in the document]
dgparmar14: The old one is the server side component and 
GithuHub Discussion is client-side comp. So the server side component will not work inside the client component, right?
So I created a new clients side comp , I know double is not needed, that's why if you approve it , I will  make the Markdwon.tsx to the client side .
I mentioned in PR description it is experimental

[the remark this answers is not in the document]
rithviknishad: You can render server component inside a client component (as long as the props are serializable)

Ref: https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#interleaving-server-and-client-components

lib/api.ts
dgparmar14: I am planning that, We have to give a points for discussions.
So I there will be three terms based on that we have to give a point
1. Comment_created in discussions
2. Discussion_created (author)
3. Who answered that discussion

So My plan is for comment_created in the discussions we can merge that with comment_created as already we have that field

image
image

components/DscMarkdown.tsx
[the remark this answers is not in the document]
rithviknishad: Also, as mentioned in other reviews, we'd be moving away from discussions being a client component by not requiring to use a state and useEffect. So this wouldn't be an issue

app/discussions/page.tsx
```diff
@@ -0,0 +1,15 @@
...
+import GithubDiscussions from "../../components/discussions/GithubDiscussions";
+
+const page = async () => {
+  const discussions = await fetchGithubDiscussion();
+
+  return (
+    <div className="gap-5">
+      <GithubDiscussions discussions={discussions} />
+      {/* In future we have more discussion here */}
+    </div>
```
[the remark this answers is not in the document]
dgparmar14: I make this "flex flex-col gap-5" Because in the future we have More components below <GithubDisucssion />. For now it is not use full

lib/types.ts
dgparmar14: Because there is no need of a text field while parsing GitHub discussion (in api.ts), However I will revert this and put
text : "", in GitHub discussion

app/discussions/page.tsx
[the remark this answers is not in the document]
rithviknishad: We can always add that `div` element back when the requirement comes. At present we haven't scoped out what more would come in the future and how it'd come right?

lib/types.ts
rithviknishad: What about the discussion body? That can be kept as the text right?

dgparmar14: When you see GithubDicussion in GithubActivity.tsx there is no need to use text. 
To render the discussion body we already pass the whole discussion in <GithubDicussion.tsx >.

components/discussions/GithubDiscussion.tsx
```diff
@@ -0,0 +1,183 @@
...
+      {!isProfilePage && (
+        <div className="relative min-w-10">
+          {/* Profile image */}
+          <Image
+            className="flex h-10 w-10 items-center justify-center rounded-full bg-secondary-400 ring-8 ring-secondary-200 transition-all duration-200 ease-in-out group-hover:scale-125 group-hover:ring-2 dark:ring-secondary-800 group-hover:dark:ring-white/50"
+            src={`https://avatars.githubusercontent.com/${discussion.author}`}
+            alt={`${discussion.author}'s profile picture`}
+            height={48}
+            width={48}
+          />
```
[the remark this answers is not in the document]
dgparmar14: In coomponetn/contrbutors we have one component ContributorImage.tsx so we can modify that component direclty and reuse it if you agree?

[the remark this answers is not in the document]
rithviknishad: Yup! I believe that also adds some shadow animation based on rank right? We could disable that logic if no rank is specified in the prop.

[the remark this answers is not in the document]
dgparmar14: And as you say that not limited to discussion only 
so I need to reuse this everywhere in project where ever contributor image use rendered right?
And I see some where &64 and &108 is used with avatar url if it's mandatory than I will create prop for this 
and on more thing is there will be class name is different at some place it is also can be handled by props ?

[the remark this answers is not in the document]
rithviknishad: Yup, I'd say `size` can be a prop that can either be one of `"sm" | "md" | "lg"` (default to "md"). The numeric size value (for query params) and classes can be derived based on this value and the `rank` prop.

This way we'd not require classNames prop to be passed.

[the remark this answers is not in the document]
rithviknishad: Feel free to create a separate PR for this change, with base branch set to `main`. We can sync that back to this PR.

lib/api.ts
dgparmar14: image

If i remove types from there than generate this error so the type declaration is require there

rithviknishad: The correct way to solve that is to have correct type definition for activity.type instead.

components/DscMarkdown.tsx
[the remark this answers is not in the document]
dgparmar14: There is still issue because as GithubDiscussion.tsx is still clientcomponent right and we have to pass the discription in the markdwon as children and description will render condtionally for readmore and readless.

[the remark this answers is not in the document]
rithviknishad: Like mentioned previously, you can interleave client and server components. Refer the link mentioned above

lib/utils.ts
```diff
@@ -88,12 +88,31 @@ export const navLinks = [
...
 export const formatDate = (date: Date) => {
   return format(date, "MMM dd, yyyy");
 };
-type Features = "Projects" | "Releases";
+type Features = "Projects" | "Releases" | "Discussions";
 export const featureIsEnabled = (feature: Features) => {
   return env.NEXT_PUBLIC_FEATURES?.split(",").includes(feature);
 };
+
+export const parseOrgRepoFromURL = (url: string) => {
```
rithviknishad, line 102:
```suggestion
export const parseOrgRepoFromURL = (url: string) : { org: string; repo: string | null } => {
```

lib/utils.ts
```diff
@@ -88,12 +88,31 @@ export const navLinks = [
...
+  const parts = url.split("/");
+
+  let org, repository;
+
+  if (parts[3] === "orgs") {
+    // Handle the URL format: /orgs/{org}/discussions/{id}
+    org = parts[4];
+    repository = "";
```
rithviknishad, line 110: It's a good practice to use `null` instead of empty string for when any information is not available.

rithviknishad (changes requested):

components/Markdown.tsx
```diff
@@ -16,7 +16,7 @@ export default async function Markdown(props: {
     .process(props.children || "");
 
   return (
-    <div className="prose font-inter text-sm leading-relaxed dark:prose-invert sm:text-base xl:text-left">
+    <div className="prose font-inter text-sm leading-relaxed dark:prose-invert prose-h2:mt-3 sm:text-base xl:text-left">
```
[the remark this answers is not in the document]
dgparmar14: Generally heading is in h1 or h2 and we have problem with only margin-top so h1 has 0 margin-top in porse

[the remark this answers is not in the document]
rithviknishad: Okay. Just this here for future reference.

https://github.com/tailwindlabs/tailwindcss/discussions/3006#discussioncomment-148487

lib/discussion.ts
```diff
@@ -0,0 +1,47 @@
...
+    return discussions[0].filter(
+      (discussion: ParsedDiscussion) =>
+        (discussion.participants ?? []).includes(user) ||
+        discussion.author === user,
+    );
+  }
+
+  return noOfDiscussion
+    ? discussions[0].slice(0, noOfDiscussion)
+    : discussions[0];
```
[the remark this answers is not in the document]
dgparmar14:
```
  const discussions = filesInDir.map((file) => {
    const content = fs.readFileSync(path.join(GH_DATA, file)).toString();
    return JSON.parse(stripJsonComments(content));
  })[0] as ParsedDiscussion[];
```

This is the right one.

lib/api.ts
```diff
@@ -115,13 +130,13 @@ export async function getContributorBySlug(file: string, detail = false) {
   };
 
   const weightedActivity = activityData.activity.reduce(
-    (acc, activity) => {
+    (acc, activity: Activity) => {
```
dgparmar14, line 133: Bro here is type mismatch i don't know why i try to solve it but not able to solve

lib/discussion.ts
```diff
@@ -10,38 +13,121 @@ const GH_DATA = join(
   process.env.DATA_REPO ?? "data-repo",
   "../data/github/discussions",
 );
-export const categoriesMap = new Map();
+export const categoriesArray: { name: string; emoji: string }[] = [];
```
rithviknishad, line 16: Just keeping it `categories` would be ideal. We know implicitly it's an array by reading it.

lib/types.ts
```diff
@@ -42,6 +43,10 @@ export interface Highlights {
   pr_collaborated: number;
   issue_assigned: number;
   issue_opened: number;
+  pr_stale?: number; // stale PRs added because there is need in app/api/leadeborad/functions.ts to avoid type error
```
rithviknishad, line 46:
```suggestion
  pr_stale?: number;
```

components/discussions/GithubDiscussions.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface Props {
+  discussions: ParsedDiscussion[];
+  searchParams?: { [key: string]: string };
+}
+
+const GithubDiscussions = ({ discussions, searchParams }: Props) => {
+  const category = searchParams?.category;
+  const dateRange = searchParams?.between;
+  const [start, end] = parseDateRangeSearchParam(dateRange);
```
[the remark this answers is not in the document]
rithviknishad: Not resolved yet...

components/discussions/GithubDiscussions.tsx
```diff
@@ -0,0 +1,51 @@
...
+    discussions = discussions.filter(
+      (discussion) => discussion.category?.name === category,
+    );
+  } else if (dateRange) {
+    discussions = discussions.filter(
+      (discussion) =>
+        new Date(discussion.time) >= new Date(start) &&
+        new Date(discussion.time) <= new Date(end),
+    );
+  }
```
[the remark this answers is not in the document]
rithviknishad: Not resolved yet!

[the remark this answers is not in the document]
dgparmar14: When you see below if else i added this code
image

and we are passing date range 30 so it will fetch for last 30 days right?

components/discussions/GithubDiscussions.tsx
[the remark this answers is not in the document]
dgparmar14: image

It is updated i keep the date range separate because it is required in if condition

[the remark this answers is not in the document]
rithviknishad: By using `parseDateRangeSearchParam` utility function, you will always have a date range. So that check can be avoided altogether...

components/discussions/GithubDiscussions.tsx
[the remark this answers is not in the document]
rithviknishad: https://github.com/coronasafe/leaderboard/pull/463#discussion_r1682709068

Also, why perform `new Date(start` and `new Date(end)`? start and end already are date objects!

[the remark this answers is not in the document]
dgparmar14: image
I hope this solve the issue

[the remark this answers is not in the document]
rithviknishad: Yup!!! This is way more readable and cleaner than before!

rithviknishad (approved):

config/GraduateAttributes.ts
```diff
@@ -290,6 +311,11 @@ export let resolveGraduateAttributes = (
         attribute,
         contributor.activityData?.authored_issue_and_pr?.length,
       );
+    case "empathy":
+      return resolveLevel(
+        attribute,
+        contributor.highlights.discussion_answered / 5,
```
rithviknishad, line 317: Why divide by 5?

lib/discussion.ts
```diff
@@ -7,10 +7,54 @@ import octokit from "./octokit";
...
+            node {
+              author {
+                login
+              }
+            }
+          }
+        }
+      }
+    }
+  }`);
```
rithviknishad, line 48: Use pagination. What if there are more then 100 comments?

scraper/src/github-scraper/discussion.ts
```diff
@@ -52,43 +44,48 @@ async function fetchDiscussionsForOrg(org: string, cursor = null) {
...
-        category: {
-          name: d.node.category.name,
-          emoji: d.node.category.emojiHTML.replace(/<\/?div>/g, ""),
-        },
-        participants,
-        repoName: repo.repoName,
-      };
-    }),
+  const endDate: Date = startOfDay(parseISO(date));
+  const startDate: Date = startOfDay(subDays(endDate, numDays));
```
rithviknishad, line 64: this is duplicated twice in the scraper. Pass the date range objects directly instead of repeating the logic. Easier to maintain

rithviknishad (changes requested):

config/GraduateAttributes.ts
dgparmar14: Highlights.discussion answered is point write ?
Now let the user have 50 points for discussion answered 
so he answered no of discissions is 50/5 10 okk?

rithviknishad: Divide by referencing that instead. What if the points is changed later on?

dgparmar14: image

rithviknishad: yup

components/Markdown.tsx
```diff
@@ -4,22 +4,26 @@ import remarkGfm from "remark-gfm";
...
-    .use(rehypeStringify)
-    .process(props.children || "");
+  const processedMarkdown = useMemo(() => {
+    return unified()
+      .use(remarkParse)
+      .use(remarkGfm)
+      .use(remarkRehype)
+      .use(rehypeStringify)
+      .processSync(props.children || ""); // Using processSync for efficiency
+  }, [props.children]);
```
rithviknishad, line 20: Why is this useMemo added? 
We only have `className` and `children` that could cause the component to re-render. useMemo does not have any effect here right? (Correct me if I'm wrong)

dgparmar14: I am finding out why the page is rendered after each filter.
So I find that there is an issue with markdown.tsx 
and in that there are two things we are using one is `unified()` and another is `dangerouslySetInnerHTML` so that is lead to re-render (I got to know by bard and ChatGPT) and they give the solution of use memo .

rithviknishad: `useMemo` is not what solved the issue here. Making the component a sync component instead of an async component is what solved the issue. Making it sync without useMemo alone solves the issue.

scraper/src/github-scraper/discussion.ts
```diff
@@ -37,54 +36,111 @@ query($org: String!, $cursor: String) {
...
+}`;
 
-async function fetchDiscussionsForOrg(org: string, cursor = null) {
-  const variables = { org, cursor };
-  const response = await octokit.graphql.paginate(query, variables);
+async function fetchGitHubDiscussions(
+  org: string,
+  endDate: Date,
+  startDate: Date,
+  cursor = null,
```
rithviknishad, line 53: Why pass cursor when using `.paginate`? This is not even being set in any of the callees.
```suggestion
```

scraper/src/github-scraper/discussion.ts
```diff
@@ -37,54 +36,111 @@ query($org: String!, $cursor: String) {
...
+) {
+  const variables = {
+    org,
+    cursor,
+  };
+  for await (const response of octokit.graphql.paginate.iterator(
+    query,
+    variables,
+  )) {
+    const repositories = await response.organization.repositories.edges;
```
rithviknishad, line 63: Why await a non async property?

```suggestion
    const repositories = response.organization.repositories.edges;
```

scraper/src/github-scraper/discussion.ts
```diff
@@ -37,54 +36,111 @@ query($org: String!, $cursor: String) {
...
+    type repo = (typeof repositories)[0];
+    for (const repo of repositories) {
+      const discussions = await repo.node.discussions.edges.map(
+        (discussion: repo) => ({
+          repoName: repo.node.name,
+          discussion: discussion.node,
+        }),
+      );
```
rithviknishad, line 71: 1. Repeating for the fourth time again... Use proper naming for types.
2. Also, why would you infer type of repo and use it for mapping discussions?
3. Are you sure repoName attribute is correct? 

```suggestion
    for (const repo of repositories) {
      const discussions = await repo.node.discussions.edges.map(
        (repo) => ({
          repoName: repo.node.name,
          discussion: repo.node,
        }),
      );
```

scraper/src/github-scraper/discussion.ts
```diff
@@ -37,54 +36,111 @@ query($org: String!, $cursor: String) {
...
+      const discussionsWithinDateRange = await discussions.find((d: repo) => {
```
rithviknishad, line 72: 1. Why await a sync call?
2. Why would discussion have a type of repo?

```suggestion
      const discussionsWithinDateRange = discussions.find((d) => {
```

scraper/src/github-scraper/discussion.ts
```diff
@@ -37,54 +36,111 @@ query($org: String!, $cursor: String) {
...
-  const discussions = response.organization.repositories.edges.map(
-    (edge: Edge) => edge.node.discussions.edges,
-  );
+        return (
+          createdAt >= new Date(startDate) || updatedAt >= new Date(startDate)
+        );
+      });
+      if (discussionsWithinDateRange) {
+        return discussions;
+      }
```
rithviknishad, line 82: Are you sure this fetches all discussions? I don't see any logic that accumulates valid discussions that passes the filter rules and return them all. This just returns as soon as the first discussions within range is valid right? What about the rest of the pages?

tests/github-discussion-schema.test.mjs
```diff
@@ -26,11 +26,13 @@ const filesInDir = fs
 
 filesInDir.forEach((file) => {
   const content = fs.readFileSync(join(GH_DATA, file)).toString();
-  const data = JSON.parse(stripJsonComments(content));
+  if (content !== "null") {
```
rithviknishad, line 29: Why check for contents of the file being "null"? When would this ever happen?

rithviknishad (changes requested):

scraper/src/github-scraper/discussion.ts
dgparmar14: Ohh i did not noticed that 

So the logic must be ``createdAt < new Date(startDate) || updatedAt < new Date(startDate)` If it true than we reach the page where the discussions are older than the start date because we are fetching based last updated and in desc order

Correct me if i am wrong

tests/github-discussion-schema.test.mjs
dgparmar14: When  storing a logic of github discussions, if there is no discussions found then empty fill is created so null indicates empty file and if we do not create empty file then this way also testing will fail .

rithviknishad: skip creating the file altogether if nothing is there to be written instead

scraper/src/github-scraper/discussion.ts
rithviknishad: It's not about the filter logic. It's about when/where/how you are using the return statement.

Does it return all the discussions within the date range or just the valid discussions from the first page alone?

dgparmar14: Got it,
Bro so need to create a new variable and whenever it is within range than append with it
and if the date range is hit than that new variable will be returned so that way we have correct array of discussion and good use of iterator

tests/github-discussion-schema.test.mjs
dgparmar14: Then also the test will be failed, won't it?

rithviknishad: `filesInDir` would result in empty array right? So why would it fail?

scraper/src/github-scraper/discussion.ts
dgparmar14: Here is the updated code 
image
I guess this will solve all the above problems all type related issues and naming conventions issues and logical issues

tests/github-discussion-schema.test.mjs
dgparmar14: So for that we have to create an empty directory?

rithviknishad: You can simply skip the whole write step altogether when there is no data. Empty directories won't get committed to git anyhow.

scraper/src/github-scraper/discussion.ts
```diff
@@ -2,9 +2,7 @@ import { octokit } from "./config.js";
 import { Discussion, ParsedDiscussion } from "./types.js";
 import { saveDiscussionData } from "./utils.js";
 
-// Query to fetch discussions from GitHub
-const query = `
-query($org: String!, $cursor: String) {
+const query = `query($org: String!, $cursor: String) {
   organization(login: $org) {
     repositories(first: 100, after: $cursor) {
```
rithviknishad, line 7: repositories also needs to be sorted by updated at desc. right? So that we ensure that discussions are always scraped for even the oldest active projects.

rithviknishad (changes requested):

scraper/src/github-scraper/discussion.ts
rithviknishad: The issue still remains same. THe logic for createdAt startDate updatedAt endDate is fine. But the issue is about: "Does it return all the discussions within the date range or just the valid discussions from the **first page alone**?"

scraper/src/github-scraper/discussion.ts
```diff
@@ -50,44 +50,42 @@ async function fetchGitHubDiscussions(
...
-  };
-  for await (const response of octokit.graphql.paginate.iterator(
-    query,
-    variables,
-  )) {
-    const repositories = await response.organization.repositories.edges;
-    type repo = (typeof repositories)[0];
+  const iterator = octokit.graphql.paginate.iterator(query, { org });
+
+  let Discussions: { repository: string; discussion: Discussion }[] = [];
```
rithviknishad, line 56: Please use proper naming conventions! `discussions`

rithviknishad (changes requested):

scraper/src/github-scraper/discussion.ts
dgparmar14: As per the above logic this is the debug view how it works please give it try once 
image

tests/github-discussion-schema.test.mjs
dgparmar14: Okk skip the whole write step but if there is no discussions directory then also test will be fail
image

rithviknishad: Run the test only if the directory is present then?

tests/github-discussion-schema.test.mjs
```diff
@@ -19,7 +19,14 @@ const schema = await yaml.parse(fs.readFileSync(SCHEMA_FILE).toString());
...
 
-// There will be more files in the folder becuase futrue plan is sharded data
+if (!fs.existsSync(GH_DATA)) {
+  console.log(
+    `Directory ${GH_DATA} does not exist. Skipping discussions test.`,
+  );
+  process.exit(0);
+}
+
+// There will be more files in the folder becuase futrue plan is sharded dat
```
rithviknishad, line 29:
```suggestion
```

scraper/src/github-scraper/discussion.ts
dgparmar14: Used `discussionsList`  for more clarity

rithviknishad: perfect

rithviknishad:
@dgparmar14 Builds failing when there is no discussion directory (which happens when no discussions were scraped so far and none were present to be scrapped in the last 24 hours or the given scraping time range)

image

rithviknishad (approved):
Code changes: LGTM ✅ 

Tested discussions locally and workflow in https://github.com/coronasafe/leaderboard-data-test/actions/runs/10089604078. Working as expected! ✅


Nice work @dgparmar14

sainak (approved):

Issue #472, opened 2024-07-31
Home page contributor ranking not consistent with the leaderboard page


rithviknishad:
### URL:

https://leaderboard-data-pynu1qx6g-ohcnetwork.vercel.app/

### Browser Name:

N/A

### Browser Version:

N/A

### Operating System:

N/A

### How to reproduce the issue:

Home page contributor ranking not consistent with the leaderboard page

Home Page: https://leaderboard-data-pynu1qx6g-ohcnetwork.vercel.app/

<img width="556" alt="image" src="https://github.com/user-attachments/assets/f6c978fc-c0bd-4d09-9098-a6e93192bbc4">

Leaderboard Page: https://leaderboard-data-pynu1qx6g-ohcnetwork.vercel.app/leaderboard

<img width="1411" alt="image" src="https://github.com/user-attachments/assets/d469b648-9dc9-45c6-be61-36f04b8de316">

Jash2606:
Hey @rithviknishad, I have raised a PR for this issue. Please check it and let me know if you need any more changes!

PR #473, opened 2024-08-01
Make Home page contributor ranking  consistent with the leaderboard page…

Jash2606:
Fixes: #472

#### Summary:
Resolved the issue where contributor rankings on the home page were not consistent with those on the leaderboard page by making necessary CSS adjustments.

---

Let me know if you need any further adjustments!


rithviknishad (changes requested):
The issue is about, the ranking logic seems to be having an issue. Not related to styling.

PR #482, opened 2024-08-07
Added Twitter icon

Indrajeety993648:
Screenshot from 2024-08-08 00-42-18

### DIFF
--- a/components/Footer.tsx
+++ b/components/Footer.tsx
@@ -1,6 +1,8 @@
 import Image from "next/image";
 import { env } from "@/env.mjs";
 import { FaYoutube, FaLinkedin, FaGithub, FaEnvelope } from "react-icons/fa";
+import { BsTwitterX } from "react-icons/bs";
+
 import { ReactNode } from "react";
 
 const SocialLink = ({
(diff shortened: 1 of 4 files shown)


Indrajeety993648:
@rithviknishad   please review  
Please review the changes. Currently, I have added the Twitter icon, which does not render as a clickable link. I have also added a hover effect similar to the other icons. Additionally, I corrected an issue related to Prettier, as version 5.5.4 does not support.

components/Footer.tsx
```diff
@@ -94,26 +95,38 @@ export default function Footer() {
...
+                    label="GitHub"
+                  />
+                }
+                {
+                  <SocialLink
+                    href={`mailto:${env.NEXT_PUBLIC_CONTACT_EMAIL}`}
+                    icon={FaEnvelope}
+                    label="Email"
+                  />
+                }
```
rithviknishad, line 125: Could you make these optional too?

components/Footer.tsx
```diff
@@ -94,26 +95,38 @@ export default function Footer() {
...
+                {/* Static Twitter icon */}
+                <div className="text-gray-300 transition-colors duration-300 hover:text-[rgb(176,142,230)]">
+                  <BsTwitterX className="h-6 w-6" aria-label="Twitter" />
+                </div>
```
rithviknishad, line 129: 1. Why is the twitter icon static? 
2. Why is there no link to it based on `env` value?
3. Let's not hardcode colors. Use colors that are already defined in the themes. This is so that colors are compatible with any theme configurations for different orgs.

components/discussions/GithubDiscussion.tsx
rithviknishad, line 1: Changes in this file has nothing to do with the relevant issue right? If this is solving a particular issue, feel free to create a new issue and a separate PR to solve it.

pnpm-lock.yaml
rithviknishad, line 1: Could you discard the changes to the lock file?

rithviknishad (changes requested):

Issue #497, opened 2024-08-15
Add Hover Effect to GitHub Logo Button in Navbar


abhinav26966:
### Enter your suggestions in details:

Currently, the GitHub logo button in the navbar does not have a hover effect, making it less apparent that it is a clickable element. 
To improve the user experience and visual clarity, we need to add a hover effect to the GitHub logo button. This effect should make the button more recognisable as interactive, aligning it visually with other buttons on the site. 

<img width="1404" alt="Screenshot 2024-08-15 at 7 16 06 PM" src="https://github.com/user-attachments/assets/9977e329-983a-401e-ac1a-b6d96588422e">

Suggestions for the hover effect could include changing the background colour, scaling the logo itself, or any other visual cue that enhances its appearance as a button.

abhinav26966:
Hey @rithviknishad, 
I want to work on this issue, can you please assign this issue to me.
It would be a great feature for the website.

abhinav26966:
Hey @rithviknishad, 
Can you please check the PR for the issue as it says about the Authorization failed and please help me with the solution for the same.
It would be really helpful for me.

abhinav26966:
Hey @rithviknishad, 
Can you please check the PR #500 for the issue as it says about the Authorization failed and please help me with the solution for the same.
It would be really helpful for me.

sankshaypusa:
Hi i have added hover effect and raised a PR [PR_RAISED](https://github.com/coronasafe/leaderboard/pull/502) Could you check on it

abhinav26966:
Hey @rithviknishad , 
Can you please check the PR #500 and tell if any changes are needed.

rithviknishad:
Closing as duplicate of #500

PR #499, opened 2024-08-15
Enhance GitHub Logo Button with Hover Effect

abhinav26966:
Solves #497 

This PR enhances the user experience by adding a hover effect to the GitHub logo button in the navbar. The hover effect increases the visibility and interactivity of the button, making it more apparent as a clickable element. Specifically, the following changes were made:

- Added a transition-transform effect to the button, which scales it slightly on hover.
- Introduced a transition-colors effect to change the logo color on hover, making it turn gray for better feedback.
- These changes ensure the button feels more dynamic and responsive, improving the overall UX.

https://github.com/user-attachments/assets/9b8606e2-f746-45ad-ab2a-2709f4dcd6f4

The new hover effect will make the button more intuitive for users, aligning it visually with interactive elements on the site.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,641 +1,405 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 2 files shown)


pnpm-lock.yaml
```diff
@@ -1,159 +1,2555 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
```
rithviknishad, line 1: Could you discard changes to the lockfile?

rithviknishad (changes requested):

pnpm-lock.yaml
abhinav26966: yep, doing it right away.

Issue #487, opened 2024-08-09
Add Hover Effect on Google Summer Of  Code 


Indrajeety993648:
### 1. What I expected to happen
When hovering over the "Google Summer of Code" section, I expected a visual hover effect to provide a better user experience by highlighting the section or providing additional visual feedback.

The hover effect is a common UI/UX feature that enhances the user's interaction with the site, making it more dynamic and engaging. Currently, there is no hover effect, which makes the site feel static and less interactive.

### 3. What I plan to do (Optional but better)
I plan to implement a simple CSS-based hover effect, such as changing the background color. This would improve the visual aesthetics and make the section more user-friendly.

## Screenshot 
![Screenshot from 2024-08-09 11-18-19](https://github.com/user-attachments/assets/67f9ed84-0d0c-4443-9714-2c043b359ba4)

### Additional Notes
Including hover effects on interactive elements like the "Google Summer of Code" section can significantly improve the user's experience by providing immediate visual feedback, making the website feel more responsive and polished.

Indrajeety993648:
@rithviknishad  please assign this issue to me .

rithviknishad:
@Indrajeety993648 feel free to take this issue up after following up on your stale PRs: https://github.com/coronasafe/leaderboard/pull/482, https://github.com/coronasafe/leaderboard/pull/481.

Indrajeety993648:
@rithviknishad  please assign this issue to me .

harshpanchal04:
Please assign this issue to me . I'd like to work upon this.

harshpanchal04:
please assign this issue to me .

abhinav26966:
Hey @rithviknishad, 
Is this issue still open? 
I would like to work on this issue, can you please assign this issue to me ?

Indrajeety993648:
@rithviknishad   Sorry for the delay, I was a little busy for a few days. Now I can take this on and solve it as soon as possible.

PR #501, opened 2024-08-15
Added hover Effect in Google Summer of Code label

Indrajeety993648:
## Fixes Issue : #487
- Added a hover effect to the "Google Summer of Code" label.
- On hover, the background color changes to yellow and the text color turns white.

**Details:**

- Updated the `ACTIVE_PROJECT_LABELS` configuration with a new `className` to include the hover effect.
- Utilized Tailwind CSS utility classes: `hover:bg-yellow-500`, `hover:text-white`.

**Before:**
- The label had a static background and text color.

**After:**
- The background now turns yellow and the text color changes to white on hover.


Indrajeety993648:
@rithviknishad  please review

PR #519, opened 2024-10-01
fix: Improve co-author parsing for pr_collaborated events

vikram-8290:
- Replaced .match() with .matchAll() to capture all Co-authored-by entries.

- Enhanced regex to handle multiple entries and formatting variations.

This resolves incorrect co-author parsing.

### DIFF
--- a/scraper/src/github-scraper/parseEvents.ts
+++ b/scraper/src/github-scraper/parseEvents.ts
@@ -50,11 +50,13 @@ async function addCollaborations(event: PullRequestEvent, eventTime: Date) {
 
     collaborators.add(authorLogin);
 
-    const coAuthors = commit.commit.message.match(
-      /Co-authored-by: (.+) <(.+)>/,
+    const coAuthors = commit.commit.message.matchAll(
+      /Co-authored-by: (.+?) <(.+?)>/g
     );
     if (coAuthors) {
-      for (const [name, email] of coAuthors) {
+      for (const coAuthor of coAuthors) {
+        const name = coAuthor[1]; // Access name from the match
+        const email = coAuthor[2]; // Access email from the match
         if (isBlacklisted(name)) {
           continue;
         }


scraper/src/github-scraper/parseEvents.ts
```diff
@@ -50,11 +50,13 @@ async function addCollaborations(event: PullRequestEvent, eventTime: Date) {
...
-    const coAuthors = commit.commit.message.match(
-      /Co-authored-by: (.+) <(.+)>/,
+    const coAuthors = commit.commit.message.matchAll(
+      /Co-authored-by: (.+?) <(.+?)>/g
     );
     if (coAuthors) {
-      for (const [name, email] of coAuthors) {
+      for (const coAuthor of coAuthors) {
+        const name = coAuthor[1]; // Access name from the match
+        const email = coAuthor[2]; // Access email from the match
```
rithviknishad, line 59: Why was this changed? This is same as before right?

```suggestion
      for (const [name, email] of coAuthors) {
```

rithviknishad (changes requested):

scraper/src/github-scraper/parseEvents.ts
vikram-8290: This is more flexible and safer if you're uncertain about how many capturing groups might be returned or if you need the full match for logging or debugging.  but i case if we assumes that each item in coAuthors is already an array of two elements (the name and the email) then we can use (const [name, email] of coAuthors)

rithviknishad: 1. Are you sure the index is right?
2. The regex pattern specifies exactly two capturing groups. How would you be "uncertain" about the number of capturing groups?

rithviknishad:
@vikram-8290 kindly link the issue with closing tags: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

scraper/src/github-scraper/parseEvents.ts
vikram-8290: So, both methods are valid, but destructuring offers slightly more readability and flexibility, especially in cases where future changes might alter the number of capturing groups. But in this specific case where we're certain about the structure, either method would work correctly.

Let me know if you'd like to stick with index-based access or refactor it for destructuring for readability!

rithviknishad: refactor

Issue #529, opened 2024-10-20
Leaderboard sort is not working


rithviknishad:
Leaderboard sort is not working. It is being sorted by points, but changing it to any other value is not working.

### Steps to reproduce

1. Go to leaderboard page
2. Change from points to any other option
3. Observe the result set is not getting updated to the new sort.

![Image](https://github.com/user-attachments/assets/17c30ac1-cc9e-41c8-9eba-3050a6504208)

programmerraja:
@rithviknishad can I work on this ?

programmerraja:
@rithviknishad can you check #532

PR #532, opened 2024-10-23
Fix (LeaderBoard) filter points bug fixed

programmerraja:
- Fix (LeaderBoard) filter points bug fixed
- Handled on points change updated the state based on applied filter
- related to #529 
Screenshot from 2024-10-23 05-52-08

### DIFF
--- a/app/leaderboard/[duration]/Leaderboard.tsx
+++ b/app/leaderboard/[duration]/Leaderboard.tsx
@@ -50,26 +50,65 @@ export default function Leaderboard(props: Props) {
 
   const search = searchParams.get("search") || "";
 
-  const roles = searchParams.get("roles")?.split(",") || [];
-  const ordering = ORDERING_OPTIONS.find(
-    (option) => option.value === searchParams.get("ordering"),
-  ) || { value: "points", text: "Points" };
-  const isReversed = searchParams.get("isReversed") === "true";
+  const [roles, setRoles] = useState<string[]>(
+    searchParams.get("roles")?.split(",") || [],
+  );
 
-  let resultSet = props.data;
+  const [ordering, setOrdering] = useState<SelectOption>(
+    ORDERING_OPTIONS.find(
+      (option) => option.value === searchParams.get("ordering"),
+    ) || { value: "points", text: "Points" },
+  );
 
-  if (roles.length) {
-    resultSet = resultSet.filter((a) => roles.includes(a.user.role));
-  }
+  const [resultSet, setResultSet] = useState(props.data);
 
-  if (isReversed) {
-    resultSet = resultSet.toReversed();
-  }
+  const [isReversed, setIsReversed] = useState<boolean>(
+    searchParams.get("isReversed") === "true",
+  );
 
+  const updateResultSet = (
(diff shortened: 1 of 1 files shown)


app/leaderboard/[duration]/Leaderboard.tsx
```diff
@@ -50,26 +50,65 @@ export default function Leaderboard(props: Props) {
...
+    ORDERING_OPTIONS.find(
+      (option) => option.value === searchParams.get("ordering"),
+    ) || { value: "points", text: "Points" },
+  );
+
+  const [resultSet, setResultSet] = useState(props.data);
+
+  const [isReversed, setIsReversed] = useState<boolean>(
+    searchParams.get("isReversed") === "true",
+  );
```
Comment (rithviknishad) [inline app/leaderboard/[duration]/Leaderboard.tsx:67]: let's not have a useState for this.

rithviknishad (changes requested):

app/leaderboard/[duration]/Leaderboard.tsx
Comment (programmerraja) [reply app/leaderboard/[duration]/Leaderboard.tsx:67]: @rithviknishad we need a usestate else when we update result set it won't get render on UI. correct me if i am wrong

Comment (rithviknishad) [reply app/leaderboard/[duration]/Leaderboard.tsx:67]: so far when we update the search / role / date range, it is getting updated right? how was that working then without the useState?

Comment (programmerraja) [reply app/leaderboard/[duration]/Leaderboard.tsx]: @rithviknishad can you check updated thanks for your insights

app/leaderboard/[duration]/Leaderboard.tsx
```diff
@@ -51,19 +51,31 @@ export default function Leaderboard(props: Props) {
...
+    resultSet = props.data.sort((a, b) => {
+      const aValue = a.highlights[ordering.value as keyof typeof a.highlights];
+      const bValue = b.highlights[ordering.value as keyof typeof a.highlights];
+
+      const aNumericValue = typeof aValue === "number" ? aValue : 0;
+      const bNumericValue = typeof bValue === "number" ? bValue : 0;
+
+      return isReversed
+        ? aNumericValue - bNumericValue
+        : bNumericValue - aNumericValue;
```
Comment (rithviknishad) [inline app/leaderboard/[duration]/Leaderboard.tsx:73]: The reverse would not work if no ordering is specified since it's part of the if block

rithviknishad (changes requested):

app/leaderboard/[duration]/Leaderboard.tsx
Comment (programmerraja) [reply app/leaderboard/[duration]/Leaderboard.tsx:73]: image

can u check now @rithviknishad

Comment (rithviknishad) [reply app/leaderboard/[duration]/Leaderboard.tsx:73]: we could simply return `aNumericValue - bNumericValue` instead.

And in another if block we could do `resultSet = resultSet.toReversed()` if `isReversed` just like before. would make the code more readable.

Comment (programmerraja) [reply app/leaderboard/[duration]/Leaderboard.tsx:73]: I did because If we do like that  the iteration will happen 2 times. in our case it only happens once. if you prefer readable code let me know i will change it

rithviknishad (approved):

Issue #554, opened 2024-12-09
Ensure scraper does not fetch bot events


rithviknishad:
https://github.com/ohcnetwork/leaderboard/blob/21a7beb26d93fd71f38d8ed7b78d86e5031d9296/scraper/src/github-scraper/fetchEvents.ts#L30-L32

![Image](https://github.com/user-attachments/assets/08891001-f058-43d9-bf7c-ab3f13cb1aa7)

AdityaJ2305:
Hey @rithviknishad , I would like to work on this issue

dharmik3504:
Hello @rithviknishad 
Can you review my PR please

PR #558, opened 2024-12-14
fix : Ensure scraper does not fetch bot events

dharmik3504:
Proposed Changes

Fix: #554

Description 

i have just filter out the [bot] slug file name while picking it  in first rendering . 

Reviewer(s)

@rithviknishad

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -84,7 +84,9 @@ let contributorSlugs: Awaited<ReturnType<typeof getContributorsSlugs>> | null =
 export async function getContributorsSlugs(): Promise<{ file: string }[]> {
   if (!contributorSlugs) {
     const files = await readdir(`${root}`);
-    contributorSlugs = files.map((file) => ({ file }));
+    contributorSlugs = files
+      .filter((slug) => !slug.includes("[bot]"))
+      .map((file) => ({ file }));
   }
   return contributorSlugs;
 }


rithviknishad (changes requested):
although this solves the side effect, this does not solve the root cause. we shouldn't be storing events of bots in the first place.

dharmik3504:
> although this solves the side effect, this does not solve the root cause. we shouldn't be storing events of bots in the first place.

do you want me to check that scenario as well ??  or you are opening a new issue for that ??

rithviknishad:
nope it's the same issue. and only check for that scenario.

data repo migration to remove bot's data will be handled separately.

dharmik3504:
hello @rithviknishad 
is there a way to run the project locally without  data-repo  data ?

rithviknishad:
nope. just run `pnpm load-data` to load data from data-repo

dharmik3504:
so the data will be pick from data-repo only (the MD file ) and i want hit any event so how i can re-produce the issue as while debugging it redirect me to same function (getContributorsSlugs) which is inside lib/api.ts .

rithviknishad:
#554 has been fixed btw:

https://github.com/ohcnetwork/leaderboard/compare/a253e309a244ee4342639182755555f90763874e...a3b377b91de86335ca457087399eef6b693873ff

PR #559, opened 2024-12-14
Fix Calendar View Issue

### LINKED ISSUE #550: Contributor profile's calendar is not showing data beyond one year
Here's an example of my profile's calendar captured on Dec 2nd 2024. However it shows data only upto Dec 2nd 2023. All other aggregate metrics in the profile are correct. Just that calendar chart is not showing data beyond 1 year.

We do have data dated back in 2022.
Image

Calendar:
Image

JavidSumra:
## Changes
- Updated the Array Size for Calendar Data According to the size of JSON Object Key Values in

### DIFF
--- a/components/contributors/ActivityCalendarGitHub.tsx
+++ b/components/contributors/ActivityCalendarGitHub.tsx
@@ -24,6 +24,7 @@ export default function ActivityCalendarGit({
   const { theme } = useTheme();
 
   const getCalendarData = (year: number) => {
+    console.log(year);
     const currentYear = year;
     let dates = [];
     let date = new Date(`01-01-${year}`);
@@ -73,7 +74,7 @@ export default function ActivityCalendarGit({
   const [year, setYear] = useState(0);
   const [isOpen, setIsOpen] = useState(false);
   const [activityData, setActivityData] = useState({});
-
+  // console.log(getCalendarData(year));
   return (
     <div className="gap-3 sm:flex">
       {isBrowser && (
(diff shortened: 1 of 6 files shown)


components/contributors/ActivityCalendarGitHub.tsx
```diff
@@ -70,50 +70,32 @@ export default function ActivityCalendarGit({
   const yearDiff = Number(new Date().getFullYear()) - getFirstContribYear();
   const yearsList = lastNYears(yearDiff);
 
-  const [year, setYear] = useState(0);
+  const [year, setYear] = useState(new Date().getFullYear());
```
rithviknishad, line 73: The reason why we kept this as 0 is, if no year is selected, we would show past 365 days data. i.e. if we are on Jan 10th, 2025, we'll show Jan 10th 2024 to Jan 10th 2025's data and not just Jan 1st 2025 to Jan 10th 2025.

rithviknishad (changes requested):

JavidSumra:
@rithviknishad can you please review this PR?

rithviknishad (changes requested):
didn't solve the issue

image

image

rithviknishad (changes requested):
this breaks other things; this is not the expected rank list for the last 7 days

image

rithviknishad:
Closing due to no activity

JavidSumra:
> Closing due to no activity

@rithviknishad, as per our discussion on Slack, we agreed to revisit this in January. However, since you're currently occupied with the Care frontend, it's not feasible. Could you please reopen it?

PR #576, opened 2024-12-25
Add: Identify First-Time PR Contributors for Last Week

### LINKED ISSUE #153: Highlight new contributors that have made their first contribution in the past 1 week in landing page

JavidSumra:
## Description

 This PR implements logic to identify first-time PR contributors whose first `pr_merged` activity occurred within the specified time frame (last week). It ensures accuracy by filtering contributors based on their activities and checks if their first contribution happened during last week.

## Screenshots / Outputs
image


rithviknishad (changes requested):
instead of directly keeping the logic here, let's add it to the contributor object itself. so that this value `isNewContributor` could be reused everywhere else when needed.

rithviknishad:
Also not a fan of the design. Let's not have `@`, seems out of context. Also placing above the name doesn't seem nice.

JavidSumra:
> instead of directly keeping the logic here, let's add it to the contributor object itself. so that this value `isNewContributor` could be reused everywhere else when needed.

Shall I add this logic in [getContributorBySlug](https://github.com/ohcnetwork/leaderboard/blob/bb09761f9b5f90809907256cceca76661fadb093/lib/api.ts#L92) function?

rithviknishad:
yup! that's the right place to be

lib/api.ts
```diff
@@ -180,6 +186,24 @@ export async function getContributorBySlug(file: string, detail = false) {
...
+      .filter((act) => act.type === "pr_merged")
+      .sort((a, b) => new Date(a.time).getTime() - new Date(b.time).getTime());
+
+    if (prMergedActivities.length === 0) return false;
+
+    const firstActivityTime = new Date(prMergedActivities[0].time);
+
+    return (
+      firstActivityTime >= lastWeekStart && firstActivityTime <= lastWeekEnd
+    );
```
rithviknishad, line 204: you can reduce the complexity by using `Array.some(...)` instead.

```suggestion
    const aWeekAgo = sub(new Date(), 7 days);
    
    return activity?.some((activity) => {
      if (activity.type !== "pr_merged") return false;
      return isAfter(activity, aWeekAgo));
    });
```

rithviknishad (changes requested):

lib/api.ts
JavidSumra: I believe this needs some adjustments. It should only count the first PR merged by each contributor, not all PRs merged by them, right @rithviknishad  ?

rithviknishad: oh yes, my bad

JavidSumra:
@rithviknishad can you please review this PR?

lib/api.ts
```diff
@@ -180,6 +181,20 @@ export async function getContributorBySlug(file: string, detail = false) {
...
+  const isNewContributor = (() => {
+    const { activity } = activityData;
+
+    const aWeekAgo = sub(new Date(), { days: 7 });
+
+    const firstActivity = activity
+      .filter((act) => act.type === "pr_merged")
+      .sort(
+        (a, b) => new Date(a.time).getTime() - new Date(b.time).getTime(),
+      )[0];
```
rithviknishad, line 193:
```suggestion
      .sort(compareAsc)[0];
```

lib/api.ts
```diff
@@ -180,6 +181,20 @@ export async function getContributorBySlug(file: string, detail = false) {
...
+
+    return isAfter(new Date(firstActivity?.time), aWeekAgo);
```
rithviknishad, line 195: if incase firstActivity is null, it ends up being equivalent to: `new Date()` and that'd always return true when doing `isAfter` right?

```suggestion
    if (!firstActivity) return false;
    return isAfter(firstActivity.time, aWeekAgo);
```

rithviknishad (changes requested):

lib/api.ts
JavidSumra: Date-fns seems pretty interesting to me, but I don't know much about it yet 😂.

app/page.tsx
```diff
@@ -26,6 +26,13 @@ export default async function Home() {
   const discussions = await fetchGithubDiscussion(5);
   const startDate = parseISO(env.NEXT_PUBLIC_ORG_START_DATE);
 
+  const topContributors = [
+    ...contributors.filter((contributor) => contributor.isNewContributor),
+    ...contributors
+      .filter((contributor) => !contributor.isNewContributor)
+      .slice(0, 8),
+  ];
```
rithviknishad, line 34: use Set instead to handle duplicates.

app/page.tsx
```diff
@@ -166,24 +173,24 @@ export default async function Home() {
                         role="list"
                         className="mt-4 space-y-4 sm:grid sm:grid-cols-2 sm:gap-6 sm:space-y-0 lg:grid-cols-2 lg:gap-8"
                       >
-                        {contributors
-                          .slice(0, 8)
-                          .map((contributor: any, index: number) => {
+                        {topContributors.map(
+                          (contributor: any, index: number) => {
```
rithviknishad, line 177: while we are at it, let's not use `any` type

lib/api.ts
```diff
@@ -180,6 +181,20 @@ export async function getContributorBySlug(file: string, detail = false) {
...
   const calendarData = getCalendarData(weightedActivity.activity);
 
+  const isNewContributor = (() => {
+    const { activity } = activityData;
+
+    const aWeekAgo = sub(new Date(), { days: 7 });
+
+    const firstActivity = activity
+      .filter((act) => act.type === "pr_merged")
+      .sort((a, b) => compareAsc(new Date(a.time), new Date(b.time)))[0];
```
rithviknishad, line 191: why wrap in date?

```suggestion
      .sort((a, b) => compareAsc(a.time, b.time))[0];
```

lib/api.ts
rithviknishad: not resolved; we don't need the null check in isAfter call right?

rithviknishad (changes requested):

lib/api.ts
JavidSumra: @rithviknishad `compareAsc` expects arguments of type `Date` or `number`, so I wrapped the value in a `Date` instead of converting it from a string to a number/Date.

app/page.tsx
JavidSumra: Could you please explain why a set is needed in this case? We're simply adding the top contributors to highlight them for the week, right? Since the isNewContributor attribute prevents duplication, isn't that sufficient?

rithviknishad: sufficient yes, but it's just cleaner than too many spread operations and filters.

JavidSumra: I believe spread operators and filters are necessary since it’s an array of objects, and sets wouldn’t be usable in this case for the filtering process, right?

JavidSumra:
@rithviknishad can you please review this PR and clarify doubts?

lib/api.ts
rithviknishad: compareAsc works with date/number/string too

image


image

app/page.tsx
```diff
@@ -26,6 +27,13 @@ export default async function Home() {
   const discussions = await fetchGithubDiscussion(5);
   const startDate = parseISO(env.NEXT_PUBLIC_ORG_START_DATE);
 
+  const topContributors = [
```
rithviknishad, line 30: use proper variable names. is this still topContributors? or highlight contributors?

rithviknishad (changes requested):

lib/api.ts
JavidSumra: image
I don't know why this error is showing to me can you please confirm version of `date-fns` @rithviknishad

rithviknishad: Yup, let's switch to latest version too

lib/api.ts
```diff
@@ -180,6 +181,20 @@ export async function getContributorBySlug(file: string, detail = false) {
...
+
+    const aWeekAgo = sub(new Date(), { days: 7 });
+
+    const firstActivity = activity
+      .filter((act) => act.type === "pr_merged")
+      .sort((a, b) => compareAsc(a.time, b.time))[0];
+
+    if (!firstActivity) return false;
+
+    return isAfter(new Date(firstActivity.time), aWeekAgo);
```
rithviknishad, line 195: i think this is the third time

```suggestion
    return isAfter(firstActivity.time, aWeekAgo);
```

rithviknishad (changes requested):

lib/api.ts
JavidSumra: Oh sorry forgot about it 😬

PR #589, opened 2024-12-31
Incorrect Year according to Week Display on Leaderboard

abhimanyurajeesh:
Since we are following the ISO Week Number Calculation, it was supposed to display as **"Week 1 of 2025"** instead of **"Week 1 of 2024"**

The function was supposed correctly identify December 31, 2024 as part of week 1 of 2025 because it falls after the last Thursday of 2024 (which is December 26, 2024). While the function accurately determines the week number, it fails to handle the year correctly.

Bug

This change ensures that both the week number and year correctly reflect the ISO week date standards, especially at the boundaries between years.


abhimanyurajeesh:
This will solve the issue.

abhimanyurajeesh:
image

rithviknishad (changes requested):
Today it should be showing week 52/53 of 2024 instead right?

image

abhimanyurajeesh:
According the ISO rule it should be **"Week 1 of 2025"** but your right, it should make sense to common man.

I have made the changes according.

### Updated Code Screenshot
image

rithviknishad:
the right answer for today is, week 53. 

isn't the current logic incorrect? shows 9 entries for week 52 when i tested your logic.

image

rithviknishad (approved):
not a fan of this complicated logic, but works.

Issue #588, opened 2024-12-31
Organisation Logo not displayed on production.


DonXavierdev:
### URL:

https://contributors.ohc.network/

### Browser Name:

Chrome

### Browser Version:

131.0.6778.205 

### Operating System:

MACOS,Sonoma,14.0 (23A344)

### How to reproduce the issue:

The footer on Leaderboard page doesn't display the logo
![Image](https://github.com/user-attachments/assets/da69fda0-45e5-4ed1-b0ef-7f42cc2f48fd)

It only gets displayed on development server
![Image](https://github.com/user-attachments/assets/3f046dbb-7e95-4879-a3ee-955f6f1ea9ca)

## Steps to Reproduce
1. Go to https://contributors.ohc.network/
2. Scroll down to see the missing image in the footer

DonXavierdev:
@rithviknishad Can you please confirm this and assign it to me

DonXavierdev:
Accidently closed

DonXavierdev:
@rithviknishad  Can you please confirm this and assign it to me

PR #592, opened 2024-12-31
Add logo.png to repository 

DonXavierdev:


DonXavierdev:
@rithviknishad i think this will solve the issue .

DonXavierdev:
Fixes #588

rithviknishad:
Late for review, but we are not supposed to git track org. specific things in this repo. Use leaderboard-data repo for OHC org. specific configurations.

DonXavierdev:
@rithviknishad ok i will gitignore logo.png in leaderboard and create api to fetch it from leaderboard-data is it a valid approach ?

PR #600, opened 2025-01-08
Fix Auto Width Shrink Issues in `discussions` and `releases` page

### LINKED ISSUE #597: Early Text Wrapping
### URL:

https://contributors.ohc.network/releases

### Browser Name:

Firefox

### Browser Version:

133.0.3 (64-bit)

### Operating System:

Windows 11

### How to reproduce the issue:

Text is wrapping unnecessarily in the `What's Changed` card on the releases page, creating redundant vertical space and making the page longer.

Image

rajku-dev:
Fixes #597 

 **BEFORE**

image
image

**AFTER**

image
image

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -130,3 +130,7 @@ html.dark {
   transform: translateX(0px);
   transform: translateY(0px);
 }
+
+.prose {
+  max-width: 95%;
+}
--- a/app/globals.css
+++ b/app/globals.css
@@ -130,7 +130,3 @@ html.dark {
   transform: translateX(0px);
   transform: translateY(0px);
 }
-
-.prose {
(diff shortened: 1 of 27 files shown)


rithviknishad (approved):

app/globals.css
```diff
@@ -130,3 +130,7 @@ html.dark {
   transform: translateX(0px);
   transform: translateY(0px);
 }
+
+.prose {
+  max-width: 95%;
+}
```
rithviknishad, line 136: apply the class directly where prose is used instead. let's not do css if it's achievable with tailwind

rithviknishad (changes requested):

rithviknishad (approved):

PR #604, opened 2025-01-11
Enable Calendar to Display Data from the Past Three Years

### LINKED ISSUE #550: Contributor profile's calendar is not showing data beyond one year

rajku-dev:
Fixes #550

https://github.com/user-attachments/assets/20a2fdb0-e4a3-4346-ae19-852eb94da9f5

### DIFF
--- a/lib/api.ts
+++ b/lib/api.ts
@@ -261,9 +261,8 @@ function getCalendarData(activity: Activity[]) {
     },
     {} as Record<string, any>,
   );
-  return [...Array(365)].map((_, i) => {
-    // Current Date - i
-    const iReverse = 365 - i;
+  return [...Array(365 * 3)].map((_, i) => {
+    const iReverse = 365 * 3 - i;
     const date = new Date(
       new Date().getTime() - iReverse * 24 * 60 * 60 * 1000,
     );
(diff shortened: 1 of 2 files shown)


rajku-dev:
@rithviknishad Please review if these changes are better,
I don't think so much refactoring is required to solve this issue, so i raised the PR.

rajku-dev:
@rithviknishad Can you please review this PR.

lib/api.ts
```diff
@@ -277,9 +277,8 @@ function getCalendarData(activity: Activity[]) {
     },
     {} as Record<string, any>,
   );
-  return [...Array(365)].map((_, i) => {
-    // Current Date - i
-    const iReverse = 365 - i;
+  return [...Array(365 * 3)].map((_, i) => {
+    const iReverse = 365 * 3 - i;
```
rithviknishad, line 281: invalid fix, hardcoding is not allowed

PR #733, opened 2026-04-05
Enhance NavHeader and ThemeSelector components with improved scroll behavior and button styles

### LINKED ISSUE #732: header flickering when transitioning from full width to floating box
this is flickering when scrolling from the top most of the page (and pretty annoying)

Image

abhimanyurajeesh:
### Description

Fixes #732

Enhance NavHeader and ThemeSelector components with improved scroll behavior and button styles

image

https://github.com/user-attachments/assets/694c029d-e06f-43a6-95e7-f1ccd37f43d4


rithviknishad (changes requested):
find the root cause of why the button is having a weird color instead of adjusting in instances.

PR #757, opened 2026-04-10
Fix issues with rendering badges and exclude hidden roles in recent achievements

rithviknishad:
image


abhimanyurajeesh (commented):
Testing

